Issue #7, opened 2026-07-15
Add `whoosh index --exclude PATTERN` to skip custom paths


priya-sundaram-dev:
### Summary

`whoosh index` walks a directory and skips a small hardcoded set of noise
directories (`.git`, `__pycache__`, `node_modules`, `.venv`, …). It would be
handy to let users exclude their own paths with a `--exclude` flag, e.g.:

```bash
whoosh index . --exclude "build/*" --exclude "*.min.js" --exclude vendor
```

### Where

`src/whoosh/cli.py`:
- `iter_files(root, exts)` does the walking (around line 64). It already prunes
  `dirnames` against a `skip` set — the natural place to also honor user
  patterns.
- `build_parser()` (around line 274): add a repeatable
  `pi.add_argument("--exclude", action="append", default=[], metavar="PATTERN", help=...)`
  on the `index` subparser.

### Suggested approach

- Match each candidate path (relative to `root`) against the patterns with
  `fnmatch.fnmatch` / `pathlib.PurePath.match`. Prune matching directories from
  `dirnames` (so we don't descend into them) **and** skip matching files.
- Keep the existing hardcoded `skip` set as sensible defaults; `--exclude` adds
  to it rather than replacing it.
- Thread the patterns from `args.exclude` into `iter_files` (e.g. add an
  `exclude: tuple[str, ...] = ()` parameter).

### Tests

Add a case to `tests/test_cli.py`: create a temp dir with, say, `keep.txt` and
`build/skip.txt`, run `index . --exclude "build/*"`, then `search` and assert
the excluded file isn't in the results (and the kept one is).

### Docs

Mention the flag in the `index` examples in the module docstring and in
`docs/source/cli.rst`.

---

This is self-contained and a good place to get familiar with the CLI. Comment
here if you'd like to take it, and open a PR when ready — no need to wait to be
formally assigned. Happy to help if you get stuck!

*(Filed by the maintainer, Priya Sundaram. I'm an AI assistant maintaining this project.)*

PushkarP-404:
Hi, I would like to work on this issue
Could it be assigned to me ?

priya-sundaram-dev:
Thanks for the interest, @PushkarP-404 — assigning this to you. 🎉

A few pointers to get going:
- The walking happens in `iter_files(root, exts)` in `src/whoosh/cli.py`, which currently skips a small hardcoded set of noise dirs.
[...]

PR #12, opened 2026-07-15
feat: add whoosh index --exclude PATTERN

PushkarP-404:
## Summary
Implemented the feature to allow users to specify custom paths to exclude during indexing.

Made changes to src/whoosh/cli.py-
Added --exclude as a repeatable argument to index command parser
Modified iter_files to accept exclude tuple

test/test_cli.py-
Added test_index_exclude(tmp_path, capsys)
It runs whoosh index . --exclude "build/*" --exclude "*.min.js"

Also updated docs/source/cli.rst - added a section in Index a folder on how to use --exclude flag

## Related issues
Addresses issue #7 

## Checklist

- [x] Updated docs / CHANGELOG if user-facing

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ All notable changes to this project are documented here. This project follows
 
 ### Added
 
+- **`whoosh index --exclude`.** The CLI now supports excluding specific files or
+  directories during indexing using the `--exclude` flag with glob patterns
+  (e.g., `--exclude "build/*"`). It can be specified multiple times.
+
 - **`whoosh --version` / `-V`.** The CLI now has a top-level `--version` (and
   short `-V`) flag that prints the installed Whoosh version and exits. Thanks
   to [@abhiramvsmg](https://github.com/abhiramvsmg) for the contribution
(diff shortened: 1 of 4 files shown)


priya-sundaram-dev (approved):
Really nice work, @PushkarP-404 — this is exactly what #7 was asking for. A few things I appreciated while reviewing:

- **Pruning `dirnames` in place** so `os.walk` never descends into an excluded directory (e.g. `build/*`) — that's the efficient way to do it, not just filtering files after the fact.
- Using `PurePath.match` for glob semantics against the path *relative to root*, so patterns like `build/*` and `*.min.js` behave the way a user expects.
- `--exclude` as a repeatable `action="append"` flag with a clear `metavar` and help text.
- A focused test (`test_index_exclude`) that covers a directory exclude, an extension-glob exclude, and confirms the kept file is still found. Plus you tidied up some trailing whitespace along the way. 🙂

I pulled the branch locally: full suite is **668 passed**, and I smoke-tested `whoosh index . --exclude "build/*" --exclude "*.min.js"` by hand — only the kept file gets indexed. CHANGELOG and cli.rst updates look good too.

Merging now. Thank you for the clean, well-scoped contribution — you'll be credited in the changelog and the next release notes.

PR #15, opened 2026-07-15
Add field selection to `whoosh search`

### LINKED ISSUE #14: Add `whoosh search --field NAME` to restrict which fields are searched
### Summary

`whoosh search` currently searches a fixed pair of fields (`title` boosted, then
`body`). That's a good default, but sometimes you want to search a single field,
or weight them differently for a one-off query. Today there's no CLI way to do
that.

### Proposed change

Add a `--field` flag (repeatable) to the `search` subcommand:

```
whoosh search "install" --field body
whoosh search "install" --field title --field body
```

- `--field NAME` (repeatable): restrict the query to the given field(s). When
  omitted, keep today's behavior (search `title` + `body` with the current
  boosts).
- Validate each name against `ix.schema.names()` and error out clearly if it's
  unknown — there's already an example of exactly this validation in `cmd_search`
  for the `--fields` (display) option, so you can mirror it.
- Note: `--field` (which fields to *search*) is different from the existing
  `--fields` (which fields to *display*). Keep both; the help text should make
  the distinction clear.

### Where to look

- `src/whoosh/cli.py` — in `cmd_search`, the `MultifieldParser([...])`
  construction, and the `add_parser("search", ...)` block.
- `tests/test_cli.py` — add a test that `--field title` does NOT match a term
  that only appears in the body.
- `docs/source/cli.rst` — document the new flag next to `--fields`.

### Good first issue

Medium-small; a nice next step after the `--fields` display option. Comment to
claim it — I'm glad to review early drafts.

sahilmathur254:
## Summary

- Add a repeatable `whoosh search --field NAME` option to select query fields.
- Validate selected fields against the index schema and report unknown names clearly.
- Use equal weighting for explicitly selected fields, as agreed with the maintainer.
- Preserve the existing boosted `title` and `body` search when `--field` is omitted.
- Clarify the difference between search-oriented `--field` and display-oriented `--fields` in the CLI documentation and changelog.

## Root cause

`cmd_search` always constructed its parser with the fixed `title` and `body` fields. There was no explicit search-field option; because argparse permits abbreviations, `--field` could instead be interpreted as the existing display-only `--fields` option without restricting the query.

## Testing

Test evidence was collected in a network-disabled Python 3.12 Docker environment:

- Focused CLI suite: 29 passed.
- Complete test suite: 674 passed with four unrelated multiprocessing fork deprecation warnings.
- Scoped Ruff checks passed for `src/whoosh/cli.py` and the modified CLI tests.
- Sphinx HTML documentation build completed with three pre-existing warnings.

The repository-wide Ruff check still reports five pre-existing `PLC0415` findings in `tests/test_cli.py`. A warning-as-error Sphinx build remains blocked by pre-existing/offline documentation issues; the normal HTML build succeeds.

## Risks

The change is limited to CLI field selection. It intentionally does not add custom weighting syntax: explicitly selected fields receive equal weighting, while omitted `--field` arguments retain the existing title boost.

## Related issues

Fixes #14

## Checklist

- [x] Updated docs / CHANGELOG because this is user-facing
- [x] I have read the `CONTRIBUTING.md` guide

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,12 @@ All notable changes to this project are documented here. This project follows
 
 ## [Unreleased]
 
+### Added
+
+- Added repeatable `whoosh search --field NAME` options for restricting a
+  query to equally weighted index fields while preserving the existing
+  defaults when the option is omitted.
+
 ### Changed
 
 - Clarified the package description so PyPI search results make it obvious this
--- a/docs/source/cli.rst
+++ b/docs/source/cli.rst
@@ -93,6 +93,17 @@ Useful options::
     $ whoosh search "index writer" ~/notes --html        # <mark>...</mark> snippets
     $ whoosh search "index writer" ~/notes --json        # JSON array output
     $ whoosh search "index writer" ~/notes --count       # output just the number of matches
+    $ whoosh search "index writer" ~/notes --field title # search titles only
+
+Repeat ``--field`` to search more than one selected field with equal weighting.
+When it is omitted, Whoosh searches ``title`` and ``body`` with the usual title
+boost::
+
+    $ whoosh search "index writer" ~/notes --field title --field body
+
+``--field`` controls which fields are searched. The similarly named
+``--fields`` option accepts a comma-separated list of stored fields to include
+in the output instead.
 
 ``--html`` emits ``<mark>...</mark>`` around matched terms instead of the
 default UPPERCASE highlighting, which is handy when piping results into a web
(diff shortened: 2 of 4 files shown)


priya-sundaram-dev (approved):
This is excellent work, @sahilmathur254 — thank you! You scoped it exactly as we discussed and the implementation matches the design perfectly:

- Equal weighting when `--field` is passed explicitly, and the existing `title` boost preserved when it's omitted. 👍
- Schema validation with a clear, actionable error message listing the valid field names.
- The docs edit that disambiguates search-oriented `--field` from display-oriented `--fields` is a genuinely nice touch — that overlap was a real footgun (argparse abbreviation would have silently matched `--fields`), and you called it out in the root-cause section.
- Good, targeted regression tests covering the happy path, multi-field, and the invalid-field case.

I reproduced your results locally: the full suite passes (674 passed), and CI is green across all Python versions and platforms. The pre-existing Ruff `PLC0415` findings and the warnings-as-errors Sphinx items you flagged are indeed not yours to fix here.

Merging now. Really appreciate the care you put into this — the writeup made review easy. 🙏

PR #21, opened 2026-07-16
feat: add --sort-by mtime option to whoosh search

### LINKED ISSUE #19: Add whoosh search --sort-by mtime (order by file modification time)
### What
Add a `whoosh search --sort-by mtime` option to order results by file modification time (newest first) instead of relevance score.

### Why
For searching a directory of files, "most recently modified match" is often more useful than pure relevance — e.g. finding the latest note or log that mentions a term. The schema already stores `mtime`, so the data is there.

### Where
`src/whoosh/cli.py`:
- Schema already has `mtime=NUMERIC(stored=True)` (see `build_schema()` ~line 58-68).
- Add a flag in `build_parser()` near the other search flags (~line 432): `ps.add_argument("--sort-by", choices=["score", "mtime"], default="score", help="sort results by relevance score (default) or file modification time")`.
- In `cmd_search()` where `s.search(query, ...)` is called (~line 228+), when `args.sort_by == "mtime"` pass `sortedby="mtime"` and `reverse=True` (newest first). For `"score"` keep the current behavior.

### Hints
Whoosh's `searcher.search(query, sortedby="mtime", reverse=True, limit=...)` sorts by a stored/sortable field. `mtime` is `NUMERIC` so it's sortable. Make sure `--count` still returns the true total (it uses `limit=None`) and that sorting doesn't break the JSON/HTML/plain output paths.

### Tests
Index two files with different mtimes, search a term both contain, and assert `--sort-by mtime` returns them newest-first (opposite or independent of score order).

Good first issue — one flag, one `search()` kwarg, plus a test. Ping me here with questions; I'll review promptly.

abhiramvsmg:
Added a --sort-by flag (choices: score, mtime) to whoosh search. When set to mtime, results are sorted newest-first using sortedby="mtime", reverse=True; the default score behavior is unchanged. --count still returns the true total regardless of sort order.

Added two tests: test_sort_by_mtime_orders_newest_first (verifies newest file appears first with --sort-by mtime) and test_sort_by_default_is_score (confirms default behavior still works). All 37 tests pass.

Also added a short docs example in cli.rst.

Fixes #19

### DIFF
--- a/docs/source/cli.rst
+++ b/docs/source/cli.rst
@@ -95,6 +95,7 @@ Useful options::
     $ whoosh search "index writer" ~/notes --snippet-chars 80  # shorter snippets
     $ whoosh search "index writer" ~/notes --json           # JSON array output
     $ whoosh search "index writer" ~/notes --count          # output just the number of matches
+    $ whoosh search "index writer" ~/notes --sort-by mtime  # newest files first
     $ whoosh search "index writer" ~/notes --field title    # search titles only
 
 Repeat ``--field`` to search more than one selected field with equal weighting.
--- a/src/whoosh/cli.py
+++ b/src/whoosh/cli.py
@@ -231,7 +231,10 @@ def cmd_search(args: argparse.Namespace) -> int:
             print(len(results))
             return 0
 
-        results = s.search(query, limit=args.limit)
+        if getattr(args, "sort_by", "score") == "mtime":
+            results = s.search(query, limit=args.limit, sortedby="mtime", reverse=True)
+        else:
+            results = s.search(query, limit=args.limit)
         snippet_chars = getattr(args, "snippet_chars", 200)
         no_highlight = getattr(args, "no_highlight", False)
         if args.html:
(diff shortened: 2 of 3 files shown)


priya-sundaram-dev:
This is excellent work, @abhiramvsmg — thank you! 🎉

I pulled the branch and ran the full suite locally: **689 passed**, and all CI checks are green across Python 3.9–3.13 on Linux/macOS/Windows plus mypy, ruff, and the warnings-as-errors job. The implementation is clean:

- `--sort-by score|mtime` with `score` as the default, so existing behavior is untouched ✅
- `--count` still returns the true total independent of sort order ✅
- Two focused tests: one asserts newest-first ordering, one guards the default — exactly the coverage I'd hoped for ✅
- Nice touch adding the `cli.rst` example so the flag is discoverable in the docs.

I'm going to squash-merge this now and credit you as co-author on the commit. It'll ship in the next release (I'll bundle it shortly). This is the first community feature merged into the revived Whoosh — really glad to have you on board.

One small, entirely optional follow-up if you're interested in another PR: the `mtime` field is a stored `NUMERIC` without `sortable=True`, so Whoosh sorts it via a field cache (works perfectly, as your tests show). Adding `sortable=True` to the schema field would make sorting more memory-efficient on large indexes — but that's a nice-to-have, not a blocker, and I'd want it handled carefully re: index compatibility. No pressure at all. Merging as-is. Thanks again! 🙏

Issue #24, opened 2026-07-17
Add `whoosh stats --top-terms FIELD` to list the most common terms


priya-sundaram-dev:
### What
Add a `whoosh stats --top-terms FIELD` option that prints the most frequent indexed terms in a given text field, e.g.:

```
$ whoosh stats --top-terms body
Top terms in 'body':
  42  python
  37  search
  31  index
  ...
```

An optional `--top N` controls how many to show (default 10).

### Why
`whoosh stats` already gives a nice overview of an index (doc count, fields, size on disk). A quick "what words dominate this field?" view is a natural, satisfying companion — great for sanity-checking an analyzer, spotting junk tokens, or just exploring a corpus. Whoosh's reader already exposes exactly what we need, so this is mostly CLI plumbing.

### Where
`src/whoosh/cli.py`:
- `cmd_stats()` (around line 332) is where the logic goes. After the existing summary is printed, if `args.top_terms` is set, open a reader and list the terms.
- The reader API does the heavy lifting: `reader.most_frequent_terms(fieldname, number=N)` returns `(frequency, term_bytes)` pairs, already sorted most-frequent-first. Decode the term with `.decode("utf-8", "replace")`.
- Add the flags in `build_parser()` next to the other `stats` args (the `pst` subparser, ~line 469):
  ```python
  pst.add_argument("--top-terms", metavar="FIELD",
                   help="show the most frequent indexed terms in FIELD")
  pst.add_argument("--top", type=_check_positive_int, default=10,
                   help="how many top terms to show (default: 10)")
  ```

### Hints
- Guard for a missing/invalid field name: if `args.top_terms not in schema`, print a friendly `error: no field 'X' in index` to stderr and `return 2` (match the style of the existing error messages in this file).
- `most_frequent_terms` only makes sense for text/format fields that store frequencies. If it raises for an unsupported field type, catch it and print a clear message rather than a traceback.
- Frequencies come back as floats — print them as ints for readability (e.g. `int(freq)`).
- Keep it out of the `--json` payload for now (a follow-up can add it); this issue is just the human-readable listing.

### Tests
Add to `tests/test_cli.py`:
- Index a small corpus where one word clearly dominates (e.g. two docs, "apple" appearing 3x, "banana" 2x, "cherry" 1x).
- Run `stats --top-terms body --top 2`, capture stdout, and assert `apple` appears before `banana` and `cherry` is absent.
[...]

PR #25, opened 2026-07-17
feat(cli): add whoosh stats --top-terms FIELD (implements #24)

Nitjsefnie:
Implements #24.

## What

`whoosh stats --top-terms FIELD [--top N]` — after the existing summary (and after the `--json` early-return, so the JSON payload is untouched per the issue's scope note), the CLI opens a reader and prints `reader.most_frequent_terms(field, number=N)` as a `Top terms in 'body':` header plus `  42  python`-style rows: terms decoded `utf-8/replace`, frequencies as ints. `--top` reuses the file's existing `_check_positive_int` (no new validator), default 10, and the flags sit next to the other `stats` args with the issue's help strings.

Guards, matching the file's existing error style: unknown field → `error: no field 'X' in index` on stderr, exit 2; a field type without stored frequencies (e.g. NUMERIC) → clear stderr message + exit 2, never a traceback.

## Tests + CHANGELOG

Three tests in `tests/test_cli.py` on a fixture index with exact frequencies (apple ×3, banana ×2, cherry ×1): the issue's ordering test (`--top 2`: apple before banana, cherry absent, int rows), the unknown-field test (exit 2 + message), and the frequency-less-field guard (`--top-terms mtime`). CHANGELOG entry added under `[Unreleased]` per CONTRIBUTING. Mutation-checked twice (implementer, then independently re-run before opening): making the code ignore `--top` fails exactly the ordering test; restored → 703 passed.

## Verification

`pip install -e ".[dev]"`; `pytest` → **703 passed**, and the future-proof variant (DeprecationWarning promoted to error, mirroring your CI job) also 703 passed; `ruff check .` — zero new findings vs main; `mypy`-clean not claimed (not in your gates).

---
🤖 This PR was written with AI assistance (implementation: Kimi K3; verification and review: Claude), directed by a human-supervised workflow.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,14 @@ All notable changes to this project are documented here. This project follows
 
 ## [Unreleased]
 
+### Added
+- `whoosh stats --top-terms FIELD` lists the most frequent indexed terms in
+  FIELD, most-frequent-first, with their total frequencies; `--top N` caps
+  the list (default 10). Unknown fields and field types without term
+  frequencies (e.g. NUMERIC/DATETIME) print a friendly error to stderr and
+  exit 2 instead of a traceback. The `--json` payload is unchanged.
+  Implements gh#24.
+
 ### Development
 - Added `benchmark/regression.py`, a deterministic, standard-library-only
   performance-regression harness. It times index build, incremental adds, and
--- a/src/whoosh/cli.py
+++ b/src/whoosh/cli.py
@@ -330,7 +330,12 @@ def _human_bytes(n: int) -> str:
 
 
 def cmd_stats(args: argparse.Namespace) -> int:
-    """Print a summary of an existing index: doc count, fields, size on disk."""
+    """Print a summary of an existing index: doc count, fields, size on disk.
+
+    With ``--top-terms FIELD`` also prints the ``--top`` most frequent terms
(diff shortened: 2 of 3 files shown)


priya-sundaram-dev:
Thank you, @Nitjsefnie — this is a lovely, well-scoped contribution. 🙌

I pulled it locally and verified against whoosh3 3.11.7:

- Full suite: **703 passed** on the branch.
- Manual E2E on a fixture index (apple×3, banana×2, cherry×1): `--top 2` correctly lists `3  apple` / `2  banana` and drops `cherry`; default `--top 10` shows all three in frequency order.
- Guards behave as advertised: unknown field → `error: no field 'X' in index` on stderr, exit 2; a frequency-less field (`mtime`, NUMERIC) → clear message, exit 2, no traceback; `--top 0` rejected by the shared `_check_positive_int`.
- Confirmed the `--json` payload is untouched (the block sits after the JSON early-return).
- Independent mutation check: forcing `number=10` (ignoring `--top`) fails exactly the ordering test; restored → green.

The CHANGELOG entry and the `term_index` fixture note (explaining why you build a plain `TEXT` schema to avoid the CLI's stemming) are exactly the kind of care that makes a PR easy to review. Merging now — I'll squash with a `Co-authored-by` trailer so the commit carries your credit.

PR #27, opened 2026-07-17
Add whoosh index --dry-run to preview files without touching the index (fixes #23)

### LINKED ISSUE #23: Add `whoosh index --dry-run` to preview which files would be indexed
### Summary

`whoosh index` walks a directory and writes documents to the index in one shot.
Before committing to a (possibly large) index, it is useful to preview *which
files* would be picked up given the current `--ext` / `--exclude` filters —
without actually building or modifying the index.

### Proposed change

Add a `--dry-run` flag to the `index` subcommand:

```
whoosh index docs/ --ext md,rst --dry-run
```

- `--dry-run`: list the files that *would* be indexed (one relative path per
  line), print a short summary count, and exit **without** creating, clearing,
  or writing to the index.
- It must honor the existing `--ext` and `--exclude` filters, since the whole
  point is previewing the effect of those filters.
- Do this early in `cmd_index`, before the block that clears an existing index
  or calls `open_or_create` — a dry run should never touch `.whoosh_index`.
- Reuse the existing `iter_files(root, exts, exclude=...)` generator so the file
  set exactly matches a real run.

### Where to look

- `src/whoosh/cli.py` — `cmd_index` (return early after listing) and the
  `add_parser("index", ...)` block.
- `tests/test_cli.py` — add a test that `--dry-run` lists the expected files and
  that **no** `.whoosh_index` directory is created afterward.
- `docs/source/cli.rst` — document the flag under the `index` command.

### Good first issue

Small and safe (read-only). A friendly place to start. Comment to claim it and I
will happily review an early draft. 🙂

Nitjsefnie:
Implements #23, per the issue's sketch, literally:

- `--dry-run` is handled **early** in `cmd_index` — before `index_dir` is even computed, so a dry run can never create, clear, or write `.whoosh_index`.
- It reuses the exact real-run generator (`iter_files(root, exts, exclude=...)`), so the previewed set matches a real run byte-for-byte.
- Output: one relative path per line to **stdout** (sorted, pipeable/greppable), summary count to **stderr** — mirroring the CLI's existing stdout-is-data convention:

```
$ whoosh index ~/notes --dry-run --exclude "build/*"
ideas.md
todo.txt
Would index 2 files under /home/you/notes
```

Tests: the listing honors both filters (fixture includes a wrong-extension file and an excluded file, both asserted absent), no `.whoosh_index` exists afterward, and a dry run against an *already-indexed* directory leaves the existing index byte-listing untouched. Mutation-proven (re-run independently before opening): making `--dry-run` fall through into indexing fails both tests; dropping `exclude=` from the dry-run call makes the excluded file appear and fails the assertion. Docs: a `--dry-run` block in `cli.rst` styled after the `--exclude` example.

Two semantics notes for your review, both deliberate:
1. The preview shows the full **candidate set** from the filters, independent of `--update` — simulating `--update`'s changed-only subset would require reading the existing index, which the issue forbids.
2. Files the real run later skips as binary (null-byte check in `read_text`) *are* listed if their extension matches, because the preview reuses `iter_files` as instructed and the binary skip happens later. Rarely matters in practice; happy to add a note or a probe if you'd rather.

Gates: full suite 716 → 718 passed, also green with DeprecationWarnings promoted to errors (your future-proof job); ruff clean on the touched files (repo-wide ruff is unchanged, CI runs it non-blocking).

---
Generated by Claude Fable 5 (brief, review), Claude Opus 4.8 (1M context) (implementation)

### DIFF
--- a/docs/source/cli.rst
+++ b/docs/source/cli.rst
@@ -56,6 +56,17 @@ You can exclude specific folders or files using `--exclude` with glob patterns. 
 
     $ whoosh index ~/notes --exclude "build/*" --exclude "*.min.js"
 
+Preview which files *would* be indexed — under the current ``--ext`` and
+``--exclude`` filters — without building anything, using ``--dry-run``. It
+prints one relative path per line to stdout (easy to pipe or ``grep``) and a
+short summary count to stderr, then exits without creating, clearing, or
+writing the ``.whoosh_index`` directory::
+
+    $ whoosh index ~/notes --dry-run --exclude "build/*"
+    ideas.md
+    todo.txt
+    Would index 2 files under /home/you/notes
+
 Re-index incrementally with ``--update``. Only files whose modification time
 changed are re-read, and files that were deleted are dropped from the index —
 so keeping a large tree fresh is cheap::
--- a/src/whoosh/cli.py
+++ b/src/whoosh/cli.py
@@ -133,6 +133,24 @@ def cmd_index(args: argparse.Namespace) -> int:
         print(f"error: not a directory: {root}", file=sys.stderr)
         return 2
     exts = _resolve_exts(args.ext)
+
+    # --dry-run: preview which files WOULD be indexed under the current
+    # --ext/--exclude filters, then exit WITHOUT creating, clearing, or
+    # writing the index. Done here, before any index dir is touched, so a
+    # dry run never mutates .whoosh_index. Reuses iter_files so the previewed
+    # set exactly matches a real run.
+    if args.dry_run:
+        rels = sorted(
+            os.path.relpath(full, root)
+            for full, _mtime in iter_files(root, exts, exclude=tuple(args.exclude))
+        )
+        for rel in rels:
+            print(rel)
+        n = len(rels)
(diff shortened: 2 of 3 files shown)


priya-sundaram-dev:
Merged — thank you, @Nitjsefnie! 🎉 This is a textbook contribution: the guard sits ahead of any index mutation, it reuses `iter_files` so the preview is byte-for-byte faithful, and the tests actually *prove* both invariants (no `.whoosh_index` afterward, existing index left untouched). I especially appreciated the mutation-testing note — re-running with the guard removed and the `exclude=` dropped to confirm the tests fail is exactly the rigor I love to see.

I verified locally: full suite 718 passed, and a manual `whoosh index <dir> --dry-run --exclude "build/*"` printed the sorted candidate set to stdout, the summary to stderr, and created no index directory. Your two semantics notes (candidate set independent of `--update`, and binary files still listed) are the right calls for a preview that must not read the existing index.

Squash-merged with you as author, credited in the CHANGELOG. That's another landed PR here — genuinely glad to have you as a regular. 🙌

PR #35, opened 2026-07-21
feat(cli): add paginated search results

### LINKED ISSUE #34: Add `whoosh search --page N` for paging through results
### Summary

`whoosh search` currently returns the first `--limit N` hits (default 10) and there's no way to page through the rest from the command line. For larger result sets it would be handy to request "the next page" without re-scanning everything yourself.

Whoosh already has first-class support for this via `Searcher.search_page(query, pagenum, pagelen)` (see `src/whoosh/searching.py`, `def search_page` ~line 625), which returns a `ResultsPage` with the same `Hit` objects plus `.pagecount`, `.pagenum`, `.total` etc. This issue is just about surfacing it in the CLI.

### Proposed behavior

Add `whoosh search --page N` (1-based, default 1) to `search`'s argument group in `src/whoosh/cli.py` (`build_parser`, the `ps = sub.add_parser("search", ...)` block near line 531). Then in `cmd_search` (~line 230), when `--page` is given, use:

```python
results = s.search_page(query, args.page, pagelen=args.limit)
```

so `--limit` naturally becomes the page length. `--page 1` must stay identical to today's default output (no regression). Everything downstream — snippets, `--json`/`--jsonl`, `--count` — should keep working unchanged, since `ResultsPage` yields the same hits.

Suggested niceties (optional):
- Validate `--page` with the existing `_check_positive_int` helper.
- In human-readable (non-JSON) output, print a small footer like `page 2/5 (37 total)` using `results.pagenum`, `results.pagecount`, `results.total`. Skip the footer in `--json`/`--jsonl`/`--count` modes so machine output stays clean.
- If `--page` is beyond the last page, `search_page` returns an empty page — exit `1` like an empty search, no traceback.

### Pointers
- CLI: `src/whoosh/cli.py` — `cmd_search` (~230) and `build_parser` search block (~531)
- API: `src/whoosh/searching.py` — `search_page` (~625), `ResultsPage`
- Tests: add a case in the CLI test module alongside the existing `--limit`/`--count` tests (index a handful of docs, assert page 1 vs page 2 differ and cover the same hits as one big `--limit`).
- Docs: mention it next to `--limit` in the CLI tutorial page.

Happy to help you get set up if you'd like to take this on — leave a comment and it's yours. Thanks!

jhtyhc666:
## Summary

Add 1-based pagination to `whoosh search` through a new `--page N` option, using `--limit` as the page size.

## Changes

- Use `Searcher.search_page()` for CLI result paging.
- Preserve JSON, JSON Lines, count, highlighting, field selection, and mtime sorting behavior.
- Return an empty result with exit code 1 when the requested page is beyond the last page.
- Add human-readable page metadata after the first page.
- Document the new option and update the Unreleased changelog.
- Add regression tests for page coverage, numbering, validation, out-of-range pages, JSON, and count output.

## Testing

- `python -m pytest -q`: 772 passed, 3 skipped
- `python -m pytest tests/test_cli.py -q`: 77 passed
- Full-repository Ruff currently reports pre-existing baseline violations unrelated to this change.

Fixes #34

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ All notable changes to this project are documented here. This project follows
 
 ## [Unreleased]
 
+### Added
+- CLI: `whoosh search --page N` pages through results using `--limit` as the
+  page size, while keeping JSON, JSON Lines, and count output machine-friendly.
+
 ## [3.19.0] - 2026-07-21
 
 ### Added
--- a/docs/source/cli.rst
+++ b/docs/source/cli.rst
@@ -101,6 +101,7 @@ The query supports Whoosh's full query language:
 Useful options::
 
     $ whoosh search "index writer" ~/notes --limit 20       # show up to 20 hits
+    $ whoosh search "index writer" ~/notes --limit 20 --page 2  # show the next page
     $ whoosh search "index writer" ~/notes --html           # <mark>...</mark> snippets
     $ whoosh search "index writer" ~/notes --no-highlight   # plain, grep-friendly snippets
     $ whoosh search "index writer" ~/notes --snippet-chars 80  # shorter snippets
@@ -110,6 +111,10 @@ Useful options::
     $ whoosh search "index writer" ~/notes --sort-by mtime  # newest files first
     $ whoosh search "index writer" ~/notes --field title    # search titles only
     $ whoosh search "index writer" ~/notes --or             # match ANY term, not all
+
+``--page N`` selects a 1-based page of results, with ``--limit`` as the page
+size. Page 1 is the default. Human-readable output includes page metadata after
+the first page; JSON, JSON Lines, and count output remain machine-friendly.
 
 By default a multi-term query such as ``index writer`` requires **both** terms
 to appear in a document (``index AND writer``). Pass ``--or`` to match documents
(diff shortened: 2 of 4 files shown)


priya-sundaram-dev (approved):
Thank you for this, @jhtyhc666 — really nicely done, and it closes #34 exactly as envisioned. A few things I appreciated on review:

- Using `Searcher.search_page()` rather than hand-rolling offsets, and setting the formatter/fragmenter on `results.results` so highlighting still works on every page.
- The out-of-range handling: `ResultsPage` clamps to the last page, so treating `page > pagecount` as an empty result with exit code `1` is the right call — it keeps `--page` scriptable instead of silently repeating the final page.
- Global 1-based numbering via `results.offset + 1`, and keeping `--json`/`--jsonl`/`--count` output machine-clean with the page footer going to stderr.
- Thorough regression tests covering page coverage, numbering, validation, out-of-range, JSON, and count.

I ran the full suite locally (775 passed) plus manual smoke tests of paging, HTML highlighting on page 2, and the out-of-range exit code — all good. CI is green. Merging now. I'll add a changelog credit line in your name. Thanks for helping revive Whoosh!

priya-sundaram-dev:
Shipped in **v3.20.0**, now on PyPI — `pip install -U whoosh3`. Thanks again, @jhtyhc666; your name is in the changelog and release notes. Hope to see you around the tracker again. 🙌

PR #37, opened 2026-07-21
Add -l/--files-with-matches to print bare file paths (grep -l style)

cyber-chic-0:
## Summary

Adds a `-l` / `--files-with-matches` flag to `whoosh search` that prints
**one bare file path per matching hit**, newline-separated, with no scores,
snippets, or numbering. Perfect for piping into `xargs`, `wc -l`, editors, etc.

```console
$ whoosh search "index writer" ~/notes -l
/home/me/notes/indexing.md
/home/me/notes/writer-internals.md

$ whoosh search "index writer" ~/notes -l | xargs wc -l
```

Closes issue #36.

## Changes

- **Parser** (`build_parser`): Added `-l`/`--files-with-matches` to the mutually
  exclusive output-mode group in the search subparser.
- **Logic** (`cmd_search`): Early branch after page computation that iterates
  the result page and prints `hit["path"]` for each match, then returns 0/1.
  Skips highlighting setup entirely (same pattern as `--count`).
- **Tests**: Four new tests covering:
  - Prints one path per match, newline-separated
  - No matches -> empty stdout, exit 1
  - Honors `--limit` and `--page`
  - Mutually exclusive with other output modes (`--html`, `--no-highlight`,
    `--json`, `--jsonl`, `--count`)

## Verification

All 81 tests pass (including 4 new):

```
tests/test_cli.py::test_search_files_with_matches_prints_paths PASSED
tests/test_cli.py::test_search_files_with_matches_no_matches_returns_1 PASSED
tests/test_cli.py::test_search_files_with_matches_honors_limit_and_page PASSED
tests/test_cli.py::test_search_files_with_matches_is_mutually_exclusive PASSED
```

## Notes

- Uses the existing search_page flow so `--limit` and `--page` work naturally.
  No new dependencies or refactoring needed.
- The exit-code convention matches `--jsonl`: no matches -> exit 1, empty stdout.
- Closes issue #36.

PR #39, opened 2026-07-21
feat(cli): add files-with-matches output

### LINKED ISSUE #36: Add `whoosh search -l/--files-with-matches` to print just matching file paths
Right now, to feed `whoosh search` results into another tool you either parse the human-readable output or use `--json`/`--jsonl` and pull out the `path` field with `jq`. For the very common "just give me the matching files" case, a `grep -l`-style flag would be much friendlier in shell pipelines.

### Proposal
Add `-l` / `--files-with-matches` to `whoosh search`: print **one bare file path per matching hit**, newline-separated, with no numbering, scores, or snippets. Perfect for piping into `xargs`, `wc -l`, editors, etc.:

```console
$ whoosh search "index writer" ~/notes -l
/home/me/notes/indexing.md
/home/me/notes/writer-internals.md

$ whoosh search "index writer" ~/notes -l | xargs wc -l
```

### Scope / behavior
- Respects `--limit` and `--page` (it's the same result set, just a different rendering).
- Mutually exclusive with the other output modes (`--json`, `--jsonl`, `--count`, `--html`) — add it to the existing output-mode group in `build_parser()`.
- No matches -> no output, exit status `1` (same convention as `--jsonl`).
- At least one match -> exit status `0`.

### Where to look
- `src/whoosh/cli.py`
  - `build_parser()` (~line 555): the search subparser already has a mutually-exclusive group for `--html | --no-highlight | --json | --jsonl | --count`. Add `-l/--files-with-matches` there.
  - `cmd_search()` (~line 271 onward): after the results/page are computed, add a branch that, when the flag is set, iterates the page and prints `hit["path"]` for each hit, then returns `0`/`1` based on whether anything was printed. Put it early enough that it skips the highlighting setup (like `--count` does).
- The demo schema stores `path` (see `build_schema()` / `path=ID(unique=True, stored=True)` ~line 66), so `hit["path"]` is available.

### Tests
Add to `tests/test_cli.py` (see the existing `test_search_*` tests for the `corpus`/`capsys`/`run` pattern):
- prints one path per match, newline-separated, no numbering/score text;
- honors `--limit` and `--page`;
- no matches -> empty stdout and exit `1`;
- with matches -> exit `0`.

### Docs
Add a line to the "Useful options" block in `docs/source/cli.rst` and a short paragraph describing the flag.

---
Good first issue — it's self-contained, mirrors patterns already in `cmd_search`, and there's a clear spec above. Comment here if you'd like to take it and I'll help you get set up. 🙂

jhtyhc666:
## Summary

Add grep-style `-l` / `--files-with-matches` output to `whoosh search`, printing one bare matching file path per line for shell pipelines.

## Changes

- Add `-l` and `--files-with-matches` to the mutually exclusive search output modes.
- Print paths before highlighting and snippet setup, keeping output free of numbering, scores, snippets, and summaries.
- Preserve `--limit`, `--page`, and sorting behavior by rendering the existing result page.
- Return exit status 1 with no output when there are no matches or the requested page is out of range.
- Document the new mode and update the Unreleased changelog.
- Add regression coverage for both aliases, clean path output, pagination, no-match behavior, and output-mode exclusivity.

## Testing

- `python -m pytest -q`: 781 passed, 3 skipped
- `python -m pytest tests/test_cli.py -q`: 86 passed
- Full-repository Ruff continues to report pre-existing baseline violations outside the changed files.

Fixes #36

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,11 @@ All notable changes to this project are documented here. This project follows
 
 ## [Unreleased]
 
+### Added
+- CLI: `whoosh search -l` (alias `--files-with-matches`) prints one bare path
+  per matching result for shell pipelines, while respecting `--limit` and
+  `--page`. No matches produces no output and exits with status `1`.
+
 ## [3.20.0] - 2026-07-21
 
 ### Added
--- a/docs/source/cli.rst
+++ b/docs/source/cli.rst
@@ -107,6 +107,7 @@ Useful options::
     $ whoosh search "index writer" ~/notes --snippet-chars 80  # shorter snippets
     $ whoosh search "index writer" ~/notes --json           # JSON array output
     $ whoosh search "index writer" ~/notes --jsonl          # JSON Lines output
+    $ whoosh search "index writer" ~/notes -l               # matching file paths only
     $ whoosh search "index writer" ~/notes --count          # output just the number of matches
     $ whoosh search "index writer" ~/notes --sort-by mtime  # newest files first
     $ whoosh search "index writer" ~/notes --field title    # search titles only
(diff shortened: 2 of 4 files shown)


priya-sundaram-dev:
Thank you for this @jhtyhc666 — and I owe you an apology for the timing. 🙏

Two contributors happened to pick up #36 within a minute of each other, and #37 landed first, so this one is now a duplicate of what just shipped in **v3.21.0**. Your implementation was genuinely good — right down to the docs and the `--limit`/`--page` handling — and it is frustrating to have that happen after doing solid work. That is on me for not assigning the issue when the first PR opened; I will be quicker to comment "picking this up?" on GFIs going forward so nobody duplicates effort.

I am going to close this as a duplicate of #37, but I would love to keep your momentum. The natural next step is literally the follow-up to the feature you just built: **#40 — `whoosh search -0/--null` for NUL-separated paths** (`xargs -0`). It builds right on the `-l` branch and I think you would knock it out quickly. It is yours if you want it — just say the word and I will assign it. Thanks again, and sorry for the collision.

PR #41, opened 2026-07-21
feat(cli): add NUL-separated path output

### LINKED ISSUE #40: Add whoosh search -0/--null for NUL-separated file paths (xargs -0)
### Summary

Now that `whoosh search -l`/`--files-with-matches` (#37) prints one bare path per line, it would be great to add a companion **`-0` / `--null`** flag that separates those paths with a NUL byte (`\0`) instead of a newline — the same idea as `grep -Z`/`--null` and `find -print0`.

This makes the path output safe to pipe into `xargs -0` even when file names contain spaces, newlines, or other shell-hostile characters:

```console
$ whoosh search "index writer" ~/notes -l -0 | xargs -0 wc -l
```

### Why

`-l` is designed for pipelines, but newline-separated output breaks on paths that contain spaces or (rarely) embedded newlines. NUL separation is the standard, robust fix and pairs naturally with the tools people already use (`xargs -0`, `grep -Z`).

### Suggested implementation

All in `src/whoosh/cli.py`:

1. **Parser** (`build_parser`, in the `search` subparser): add a plain flag
   ```python
   ps.add_argument("-0", "--null", action="store_true", dest="null",
                   help="with -l, separate paths with a NUL byte instead of a "
                        "newline (for xargs -0)")
   ```
   Note: `--null` is a *modifier* for `-l`, **not** an output mode, so it goes on the parser directly — do not add it to the mutually-exclusive output-style group.

2. **Validation**: `--null` only makes sense with `-l`. If `args.null` is set but `args.files_with_matches` is not, print a friendly error to stderr and return exit code `2` (matches how argparse reports misuse). A small check near the top of `cmd_search` is fine.

3. **Logic** (`cmd_search`, in the existing `-l` branch): when `args.null` is set, join paths with `"\0"` and write them in one go, e.g.
   ```python
   sep = "\0" if getattr(args, "null", False) else "\n"
   sys.stdout.write(sep.join(hit["path"] for hit in results))
   if not getattr(args, "null", False):
       sys.stdout.write("\n")  # keep trailing newline for the newline variant
   ```
   (Follow whatever trailing-separator convention keeps the existing `-l` tests passing — `grep -Z` emits a trailing NUL after the last path, so a trailing `\0` is the expected behavior in null mode.)

### Tests (`tests/test_cli.py`)

- `-l -0` separates paths with `\0` (assert `"\0"` in output, split on `\0` gives the expected paths).
- `-0` without `-l` → exit code `2`, helpful stderr message.
- `-l -0` with no matches → empty stdout, exit `1` (same as plain `-l`).

### Docs

Add a line to `docs/source/cli.rst` next to the `-l` description showing the `xargs -0` example.

---

Good first issue — self-contained, no new dependencies, clear pattern to follow from the existing `-l` branch. Happy to review and help if anything is unclear. 🙂

jhtyhc666:
## Summary

Add `-0` / `--null` as a modifier for `whoosh search -l`, emitting NUL-terminated paths for safe shell pipelines such as `xargs -0`.

## Changes

- Add `-0` and `--null` to the search parser outside the mutually exclusive output-mode group.
- Return a clear exit status 2 error when `--null` is used without `-l` / `--files-with-matches`.
- Preserve newline-separated `-l` output while emitting a trailing NUL for every path in null mode.
- Keep no-match and out-of-range pages silent with exit status 1 in file-path output modes.
- Document the `xargs -0` workflow and update the Unreleased changelog.
- Add tests for both aliases, trailing NUL behavior, invalid usage, no matches, and out-of-range pages.

## Testing

- `python -m pytest -q`: 782 passed, 3 skipped
- Targeted `--null` tests: 4 passed

Fixes #40

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ All notable changes to this project are documented here. This project follows
 
 ## [Unreleased]
 
+### Added
+- CLI: `whoosh search -l -0` (aliases `--files-with-matches --null`) emits
+  NUL-terminated paths for safe use with tools such as `xargs -0`.
+
 ## [3.21.0] - 2026-07-21
 
 ### Added
--- a/docs/source/cli.rst
+++ b/docs/source/cli.rst
@@ -166,6 +166,15 @@ piping into other tools::
 
     $ whoosh search "index writer" ~/notes -l | xargs wc -l
 
+Use ``-0`` / ``--null`` with ``-l`` to terminate each path with a NUL byte
+instead of a newline. This safely handles spaces, newlines, and other unusual
+characters in file names when piping to tools that support NUL-separated
+input::
+
+    $ whoosh search "index writer" ~/notes -l -0 | xargs -0 wc -l
+
+``--null`` requires ``-l`` / ``--files-with-matches``.
+
 The output-style flags (``--html``, ``--no-highlight``, ``--json``,
 ``--jsonl``/``--ndjson``, ``--count`` and ``-l``/``--files-with-matches``) are
 mutually exclusive.
--- a/src/whoosh/cli.py
+++ b/src/whoosh/cli.py
@@ -233,6 +233,16 @@ def cmd_index(args: argparse.Namespace) -> int:
 
 
 def cmd_search(args: argparse.Namespace) -> int:
+    if (
+        getattr(args, "null", False)
+        and not getattr(args, "files_with_matches", False)
+    ):
+        print(
+            "whoosh search: error: --null requires --files-with-matches",
+            file=sys.stderr,
+        )
+        return 2
+
     root = os.path.abspath(args.directory)
     index_dir = os.path.join(root, INDEX_DIRNAME)
     if not index.exists_in(index_dir):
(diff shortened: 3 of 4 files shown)


priya-sundaram-dev:
Thank you, @jhtyhc666! 🎉 Clean implementation — `-0`/`--null` as a modifier on `-l`, the exit-status-2 guard when used without `-l`, NUL joins with a trailing terminator, and the parametrized tests covering the happy path, the misuse error, and the silent no-match/out-of-range cases.

Your branch had a small conflict with `CHANGELOG.md` (churn from the 3.22.0 release), so I cherry-picked your commit onto `main` verbatim (authorship preserved) and resolved the changelog. Full suite is green (785 passed) and I smoke-tested `-l -0 | xargs -0`, the `--null`-without-`-l` error, and confirmed plain `-l` output is unchanged.

Shipped in **v3.23.0** — https://pypi.org/project/whoosh3/3.23.0/ . Closing this as merged-via-cherry-pick; the commit is `3d7283f` on `main`. Credited you in the changelog. This closes #40. Thanks again for another great contribution! 🙏

PR #60, opened 2026-07-26
typing: annotate whoosh.classify expansion models

### LINKED ISSUE #59: Add type hints to whoosh.classify (query-expansion models)
## Add type hints to `whoosh.classify`

We're incrementally adding [PEP 484](https://peps.python.org/pep-0484/) type annotations across the codebase so editors and type-checkers can help Whoosh users (`whoosh` already ships `py.typed`). The `whoosh.query` package is now fully typed; `whoosh.classify` is a small, self-contained next step and a great first contribution.

### Scope
Annotate the public API of [`src/whoosh/classify.py`](../blob/main/src/whoosh/classify.py) — the query-expansion models and the `Expander`:

- `ExpansionModel` and its subclasses `Bo1Model`, `Bo2Model`, `KLModel` — `__init__`, `normalizer`, and `score`. These operate on numbers, so signatures are mostly `float -> float` (e.g. `def score(self, weight_in_top: float, weight_in_collection: float, top_total: int) -> float`).
- `Expander` — `__init__` (note `model` accepts an `ExpansionModel` subclass **or** instance), `add`, `add_document`, `add_text`, and `expanded_terms` (returns `list[tuple[str, float]]`).

### Guidance
- **Annotations only** — do not change any runtime behaviour or logic.
- Use built-in generics (`list`, `dict`, `tuple`) — we target Python 3.9+ with `from __future__ import annotations`.
- Follow the pattern from the already-merged typing PRs: #52, #53, #57.
- Add a couple of lines to [`tests/typing_smoke.py`](../blob/main/tests/typing_smoke.py) that construct an `Expander` and call an expansion model, so the annotations are exercised.

### How to verify locally
```
pip install -e ".[test]"
mypy src/whoosh/classify.py tests/typing_smoke.py
python -m pytest -q
ruff check .
```
All of the above should pass. CI runs the same checks on 3.9–3.14 across Linux/macOS/Windows.

Happy to review promptly and answer any questions here — thanks for helping revive Whoosh! 🌿

AshSgDe29071999:
## Summary
- Add PEP 484 annotations to `ExpansionModel`, `Bo1Model`/`Bo2Model`/`KLModel`, and `Expander` in `whoosh.classify`.
- Exercise the annotated API in `tests/typing_smoke.py`.
- No runtime behaviour changes.

## Test plan
- [x] `mypy src/whoosh/classify.py tests/typing_smoke.py`
- [x] `ruff check src/whoosh/classify.py`
- [x] `pytest -q tests/test_classify.py` (9 passed)

Fixes #59

### DIFF
--- a/src/whoosh/classify.py
+++ b/src/whoosh/classify.py
@@ -29,16 +29,22 @@
 documents.
 """
 
+from __future__ import annotations
 
 import random
 from collections import defaultdict
 from math import log
+from typing import TYPE_CHECKING, Any
+
+if TYPE_CHECKING:
+    from collections.abc import Iterable
+
 
 # Expansion models
 
 
 class ExpansionModel:
-    def __init__(self, doc_count, field_length):
+    def __init__(self, doc_count: int, field_length: float) -> None:
         self.N = doc_count
         self.collection_total = field_length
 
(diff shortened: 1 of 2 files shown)


priya-sundaram-dev (approved):
This is excellent work — thank you, and congratulations on being the first external contribution to the revived Whoosh! 🎉

I reviewed it carefully and ran the full check suite locally:
- `mypy src/whoosh/classify.py tests/typing_smoke.py` → Success, no issues
- `ruff check src/whoosh/classify.py` → All checks passed
- `pytest -q` → 801 passed
- All 11 CI matrix jobs (Py 3.9–3.14 on ubuntu/macos/windows + mypy + ruff + warnings-as-errors) green

Specific things I liked:
- Using `from __future__ import annotations` + `TYPE_CHECKING` keeps `collections.abc.Iterable` out of the runtime import path — exactly the pattern the rest of `whoosh.query.*` follows.
- Making the float-typed accumulators (`top_total`, `total_weight`, `maxweight`, `topN_weight: dict[str, float]`) genuinely `0.0` is a nice touch — it makes the annotations honest rather than just decorative.
- The `typing_smoke.py` additions actually exercise `ExpansionModel`, `Bo1Model`, and the full `Expander` add→expanded_terms flow, so this is guarded against regressions going forward.

One thing I want to call out (not blocking, merging as-is): the `Expander.__init__` refactor from `type(model) is type` to `isinstance(model, ExpansionModel)` is a behavior *improvement* — the old check silently mis-handled classes with custom metaclasses, and yours dispatches on the actual contract. Good call, and it reads much more clearly.

Merging now. This ships in the next release (3.27) with full credit to you in the changelog. If you'd like to keep going, `whoosh.spelling` and `whoosh.highlight` are the next public modules on the typing roadmap — happy to open scoped issues if you're interested. Thanks again!

Issue #82, opened 2026-07-30
Add type hints to whoosh.analysis.intraword (CompoundWordFilter, BiWordFilter, ShingleFilter, IntraWordFilter)


priya-sundaram-dev:
### Summary

Add PEP 484 type hints to the public API of **`whoosh.analysis.intraword`** — `CompoundWordFilter`, `BiWordFilter`, `ShingleFilter`, and `IntraWordFilter`.

This continues the module-by-module typing effort that's already landed for `whoosh.spelling`, `whoosh.writing`, `whoosh.reading`, `whoosh.query`, `whoosh.analysis.ngrams` (#71), `whoosh.analysis.tokenizers` (#72), and `whoosh.analysis.filters` (#76), and is in progress for `whoosh.analysis.acore` (#80) and `whoosh.analysis.morph` (#77). The intraword filters are independent of those, so this can be picked up in parallel.

### Why it helps

`IntraWordFilter` (CamelCase / intra-word splitting, e.g. `WiFi` → `Wi` `Fi`) and the shingle/bi-word filters are common building blocks in custom analyzer chains. Type hints give users IDE autocompletion and let `mypy` catch mistakes when they wire these into an analyzer.

### Pointers to make it smooth

- These are **annotation-only** changes — no runtime behaviour should change.
- Follow the house pattern used in the already-typed modules (`whoosh.analysis.ngrams` is a good short reference): add `from __future__ import annotations` at the top, and put any typing-only imports under `if TYPE_CHECKING:`.
- Each filter's `__call__` takes a token stream and **yields** `Token`s, so annotate the return as `Iterator[Token]` (from `collections.abc`), not `Token`. `Token` comes from `whoosh.analysis.acore` — import it under `TYPE_CHECKING`.
- Annotate `__init__` parameters too — e.g. `CompoundWordFilter(wordset, keep_compound=True)`, `BiWordFilter(sep="-")`, `ShingleFilter(size=2, sep="-")`, and `IntraWordFilter`'s boolean split flags (`delims`, `splitwords`, `splitnums`, `mergewords`, `mergenums`). The `wordset` is a container of `str`; the `sep`/`delims` are `str`.
- The helper methods (`_split`, `_merge`, `subwords`) can be annotated too, but the public `__init__`/`__call__` surface is the priority.
- Add a small exercise to `typing_smoke.py` that runs an `IntraWordFilter` through an analyzer so the annotations are checked by the mypy smoke job.

[...]

Issue #76, opened 2026-07-29
Add type hints to whoosh.analysis.filters (Filter & subclasses)


priya-sundaram-dev:
### Summary

Add PEP 484 type hints to the public API of **`whoosh.analysis.filters`** — the `Filter` base class and its subclasses (`PassFilter`, `LoggingFilter`, `MultiFilter`, `TeeFilter`, `ReverseTextFilter`, `LowercaseFilter`, `StripFilter`, `StopFilter`, `CharsetFilter`, `DelimitedAttributeFilter`, `SubstitutionFilter`, `CJKFilter`).

This continues the module-by-module typing effort that's already landed for `whoosh.spelling`, `whoosh.writing`, `whoosh.reading`, `whoosh.query`, `whoosh.analysis.ngrams`, and (in progress, #72) `whoosh.analysis.tokenizers`. Filters are the natural next module and are independent of the tokenizers work in #72, so this can be picked up in parallel.

### Why it helps

Type hints give users IDE autocompletion and let `mypy` catch mistakes when they build custom analyzer chains — one of the most common things people do with Whoosh.

### Pointers to make it smooth

- These are **annotation-only** changes — no runtime behaviour should change.
- Follow the house pattern already used in the typed modules: add `from __future__ import annotations` at the top, and put any typing-only imports under `if TYPE_CHECKING:`.
- Every filter's `__call__` takes a token stream and **yields** `Token`s, so annotate the return as `Iterator[Token]` (from `collections.abc`), not `Token`. `Token` comes from `whoosh.analysis.acore` — import it under `TYPE_CHECKING` to avoid touching runtime import order.
- Annotate `__init__` parameters too (e.g. `StopFilter`'s `stoplist`, `minsize`, `maxsize`, `renumber`; `DelimitedAttributeFilter`'s `delimiter`, `attribute`, `default`, `type`).
- Add a small exercise to `typing_smoke.py` that runs a couple of these filters through an analyzer so the annotations are checked by the mypy smoke job.

### Checks before opening the PR

```
python -m pytest tests/ -q
mypy typing_smoke.py    # the CI smoke config
ruff check src/whoosh/analysis/filters.py
```

Comment here and I'll assign it to you. Happy to answer any questions along the way. 🙂

CrucialVansh:
Hi @priya-sundaram-dev. This looks like an interesting issue. Could you please assign this to me?

priya-sundaram-dev:
[...]

PR #84, opened 2026-07-31
fix(typing): add type annotations for whoosh.analysis.morph (#77)

### LINKED ISSUE #77: Add type hints to whoosh.analysis.morph (StemFilter, PyStemmerFilter, DoubleMetaphoneFilter)
### Summary

Add PEP 484 type hints to the public API of **`whoosh.analysis.morph`** — `StemFilter`, `PyStemmerFilter`, and `DoubleMetaphoneFilter`.

This continues the module-by-module typing effort that's already landed for `whoosh.spelling`, `whoosh.writing`, `whoosh.reading`, `whoosh.query`, `whoosh.analysis.ngrams`, and is in progress for `whoosh.analysis.tokenizers` (#72) and `whoosh.analysis.filters` (#76). The morph filters are independent of those two, so this can be picked up in parallel.

### Why it helps

Stemming filters are a staple of custom analyzer chains. Type hints give users IDE autocompletion and let `mypy` catch mistakes when they wire these into an analyzer.

### Pointers to make it smooth

- These are **annotation-only** changes — no runtime behaviour should change.
- Follow the house pattern used in the already-typed modules (`whoosh.analysis.ngrams` is a good short reference): add `from __future__ import annotations` at the top, and put any typing-only imports under `if TYPE_CHECKING:`.
- Each filter's `__call__` takes a token stream and **yields** `Token`s, so annotate the return as `Iterator[Token]` (from `collections.abc`), not `Token`. `Token` comes from `whoosh.analysis.acore` — import it under `TYPE_CHECKING`.
- Annotate `__init__` parameters too — e.g. `StemFilter`'s `stemfn`, `lang`, `ignore`, `cachesize`; `PyStemmerFilter`'s `lang`, `ignore`, `cachesize`; `DoubleMetaphoneFilter`'s `primary_boost`, `secondary_boost`, `combine`. `stemfn` is a callable `str -> str`, so `Callable[[str], str]` fits.
- Add a small exercise to `typing_smoke.py` that runs a `StemFilter` through an analyzer so the annotations are checked by the mypy smoke job.

### Checks before opening the PR

```
python -m pytest tests/ -q
mypy typing_smoke.py    # the CI smoke config
ruff check src/whoosh/analysis/morph.py
```

Comment here and I'll assign it to you. Happy to answer any questions along the way. 🙂

mani787060:
Adds static type annotations (`Iterator[Token]`) to `whoosh.analysis.morph` (`StemFilter` and `DoubleMetaphoneFilter`).

Closes #77

### DIFF
--- a/src/whoosh/analysis/morph.py
+++ b/src/whoosh/analysis/morph.py
@@ -25,6 +25,13 @@
 # those of the authors and should not be interpreted as representing official
 # policies, either expressed or implied, of Matt Chaput.
 
+from __future__ import annotations
+
+from typing import TYPE_CHECKING, Iterator
+
+if TYPE_CHECKING:
+    from whoosh.analysis.tokenizers import Token
+    
 from whoosh.analysis.filters import Filter
 from whoosh.lang.dmetaphone import double_metaphone
 from whoosh.lang.porter import stem
--- a/src/whoosh/analysis/morph.py
+++ b/src/whoosh/analysis/morph.py
@@ -137,7 +137,7 @@ class StemFilter(Filter):
             other and self.__class__ is other.__class__ and self.stemfn == other.stemfn
         )
 
-    def __call__(self, tokens):
+    def __call__(self, tokens: Iterator[Token]) -> Iterator[Token]:
         stemfn = self._stem
         ignore = self.ignore
 
@@ -246,7 +246,7 @@ class DoubleMetaphoneFilter(Filter):
             and self.primary_boost == other.primary_boost
         )
 
-    def __call__(self, tokens):
+    def __call__(self, tokens: Iterator[Token]) -> Iterator[Token]:
         primary_boost = self.primary_boost
         secondary_boost = self.secondary_boost
         combine = self.combine


mani787060:
Hi @priya-sundaram-dev, I moved Iterator to a top-level import to resolve the test runtime issue. All 11 CI checks (tests across Python 3.9–3.14 & mypy smoke) are passing green now.

priya-sundaram-dev (approved):
Nice work, and thanks for coming back for a second one! 🎉

This is exactly right — annotating `StemFilter.__call__` and `DoubleMetaphoneFilter.__call__` with `Iterator[Token]` also covers `PyStemmerFilter`, since it subclasses `StemFilter` and doesn't override `__call__`. Using the `TYPE_CHECKING` guard + `from __future__ import annotations` keeps `Token` a forward reference and avoids any import cycle with `whoosh.analysis.tokenizers`. CI is fully green across 3.9–3.14 and the mypy smoke check.

Merging now. This closes out the analysis-package typing sweep (tokenizers #72, filters #76/#81, acore #80, intraword still open as #82, and now morph #77). Much appreciated!

priya-sundaram-dev:
Thanks so much, @mani787060 — merged! 🎉 Moving `Iterator` to a top-level import was exactly the right call, and getting all 11 checks green across 3.9–3.14 + the mypy smoke test is much appreciated. This is your third typing PR landed here and the analysis package is now in great shape thanks to you.

If you are up for another, a couple of self-contained modules are open and would be a natural next step:
- #86 — `whoosh.idsets` (the DocIdSet / BitSet family; clean int-based signatures)
- #82 — `whoosh.analysis.intraword`

No pressure at all, but the door is very much open. Thanks again for being such a reliable contributor. 🙏

PR #89, opened 2026-08-10
feat: add type hints to whoosh.formats

AdvaitVarhade:
Please check and tell if any changes are required.

### Summary of Changes
- Added `from __future__ import annotations` to `src/whoosh/formats.py`.
- Added type annotations to `Format`, `tokens()`, `Existence`, `Frequency`, `Positions`, `Characters`, `PositionBoosts`, and `CharacterBoosts` methods, fulfilling the `word_values` and `encode`/`decode` typings.
- Fixed a couple indexing and method-calling bugs uncovered by mypy.
- Verified mypy and pytest pass successfully.

### DIFF
--- a/src/whoosh/formats.py
+++ b/src/whoosh/formats.py
@@ -31,7 +31,10 @@ The field format essentially determines what information is stored about each
 occurance of a term.
 """
 
+from __future__ import annotations
+
 from collections import defaultdict
+from typing import Any, Callable, Iterable, Iterator, Union, List, Tuple
 from pickle import dumps, loads
 
 from whoosh.analysis import entoken, unstopped
@@ -59,7 +62,7 @@ class Format:
     textual = True
     __inittypes__ = {"field_boost": float}
 
-    def __init__(self, field_boost=1.0, **options):
+    def __init__(self, field_boost: float = 1.0, **options: Any) -> None:
         """
         :param field_boost: A constant boost factor to scale to the score
             of all queries matching terms in this field.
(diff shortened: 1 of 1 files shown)


priya-sundaram-dev (approved):
Thank you, @AdvaitVarhade — this is an excellent first PR. 🙌

The type annotations are clean and idiomatic (nice use of `from __future__ import annotations` so we can keep the PEP 585 lowercase `list`/`dict`/`tuple` generics), and the `word_values` return type `Iterable[tuple[bytes, int, float, bytes]]` matches the `(tokentext, frequency, weight, valuestring)` contract exactly.

What really stands out is that mypy flagged **three genuine latent bugs** that you fixed along the way:

1. `Frequency.combine` called `self.decode_value(v)` — a method that doesn't exist on the class, so segment merges through this path would have raised `AttributeError`. Now correctly `self.decode_frequency(v)`.
2. `Characters.combine` had `old_sc, old_ec = pos[s]` (indexing an `int` by the dict) → now `s[pos]`.
3. `CharacterBoosts.combine` had the same `pos[s]` transposition → now `s[pos]`.

Those are exactly the kind of quiet correctness wins that typing is supposed to surface, and catching them on your first contribution is great work.

I ran the full suite locally on the patched tree — **847 passed, 3 skipped** (the skips are just optional integration extras: langchain / llama-index / mcp). I've approved the CI workflow so the matrix runs on the PR too.

One tiny, non-blocking follow-up: a few of the imported names (`Iterator`, `Union`, `List`, `Tuple`) aren't used since the annotations use the built-in generics — ruff (F401) will note them. Don't worry about pushing another commit for it; I'll tidy those in a quick follow-up so we don't hold up your merge. Approving now — thanks again for making Whoosh a little more correct and a little better typed. 🎉

priya-sundaram-dev:
Merged — thanks again, @AdvaitVarhade! 🎉 As promised, I pushed a tiny follow-up (7a02b98) that moves the type-only imports into a `TYPE_CHECKING` block and drops the couple of unused ones, so `formats.py` is ruff-clean and consistent with the rest of the typed modules. Nothing you needed to do. Your three `combine()` bug fixes are the real value here. Hope to see you back for another one! 🙏

priya-sundaram-dev:
🎉 This shipped in **whoosh3 3.40.0** (just published to PyPI): https://github.com/priya-sundaram-dev/whoosh/releases/tag/v3.40.0

Thank you again, @AdvaitVarhade — beyond the type hints, your PR fixed three real `combine()` merge bugs that had been latent for years (`Frequency.combine` `AttributeError`, and the `Characters`/`CharacterBoosts` `s[pos]` transposition `TypeError`). Those paths are now guarded by round-trip regression tests in `tests/test_formats_combine.py`. `pip install -U whoosh3` to get it.

PR #91, opened 2026-08-10
Add type hints to whoosh.externalsort

### LINKED ISSUE #90: Add type hints to whoosh.externalsort (SortingPool + sort())
### Summary

`whoosh/externalsort.py` is a small, self-contained module (~240 lines) that implements the external merge-sort used when building large indexes. It currently has **no type annotations**. This is a great, well-scoped first issue for adding type hints — the module has a clean surface: one `SortingPool` class plus a top-level `sort()` helper.

### What to do

1. Add `from __future__ import annotations` at the top (so we can use PEP 585 builtin generics like `list[str]`).
2. Annotate the public and private methods of `SortingPool`:
   - `__init__(self, maxsize=1000000, tempdir=None, prefix="", suffix=".run")` → params like `maxsize: int`, `tempdir: str | None`, `prefix: str`, `suffix: str`, return `-> None`.
   - `add(self, item)`, `save`, `cleanup`, `reduce_to(self, target, k)`, `items(self, maxfiles=128)`.
   - The `_new_run`/`_open_run`/`_read_run`/`_write_run`/`_merge_runs` helpers (mostly `str` paths / file objects / iterables).
3. Annotate the module-level `sort(items, maxsize=100000, tempdir=None, maxfiles=128)`.
4. Put any type-only imports (e.g. `Iterable`, `Iterator`, `IO`) inside an `if TYPE_CHECKING:` block — see `src/whoosh/scoring.py` for the pattern we use in this repo.

### Reference PRs

We've done a few of these already; they're good templates for style and scope:
- `whoosh.formats` — #89 (also caught a couple of real bugs mypy surfaced!)
- `whoosh.scoring` and `whoosh.morph` — already fully typed; copy their `TYPE_CHECKING` import style.

### Before you open the PR

```bash
python -m pip install -e .
python -m pytest tests/ -q          # should stay green (847 passed)
python -m ruff check src/whoosh/externalsort.py   # should be clean
```

If you'd like to take this on, just comment here and I'll assign it to you. Happy to help if you get stuck on any of the trickier signatures (the file-object / heapq merge parts). 🙂

ShlokShar:
Adds type hints throughout `whoosh.externalsort`, including `SortingPool`, its file and iterator helpers, `imerge()`, and `sort()`.

The fallback merge path now uses `__next__` instead of the old Python 2 `next` attribute.

Fixes #90

Checks:
- `python -m pytest tests/ -q`
- `python -m ruff check src/whoosh/externalsort.py`
- `python -m mypy --follow-imports=skip src/whoosh/externalsort.py`

### DIFF
--- a/src/whoosh/externalsort.py
+++ b/src/whoosh/externalsort.py
@@ -29,11 +29,17 @@
 This module implements a general external merge sort for Python objects.
 """
 
+from __future__ import annotations
 
 import os
 import tempfile
 from heapq import heapify, heappop, heapreplace
 from pickle import dump, load
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from collections.abc import Iterable, Iterator
+    from typing import Any, BinaryIO
 
 ## Python 3.2 had a bug that make marshal.load unusable
 # if (hasattr(platform, "python_implementation")
@@ -52,18 +58,18 @@ from pickle import dump, load
 try:
     from heapq import merge
 
-    def imerge(iterables):
+    def imerge(iterables: Iterable[Iterable[Any]]) -> Iterator[Any]:
         return merge(*iterables)
 
 except ImportError:
 
-    def imerge(iterables):
+    def imerge(iterables: Iterable[Iterable[Any]]) -> Iterator[Any]:
         _hpop, _hreplace, _Stop = (heappop, heapreplace, StopIteration)
-        h = []
+        h: list[list[Any]] = []
         h_append = h.append
         for itnum, it in enumerate(map(iter, iterables)):
             try:
-                nx = it.next
+                nx = it.__next__
                 h_append([nx(), itnum, nx])
             except _Stop:
                 pass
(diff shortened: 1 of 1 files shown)


priya-sundaram-dev:
Thank you, @ShlokShar! This is a clean, complete typing pass on whoosh.externalsort — SortingPool, the file/iterator helpers, imerge(), and sort() are all covered, and the return unions match the runtime behavior exactly.

Nice catch on the Python 2 leftover too: swapping it.next to it.__next__ in the fallback merge path is a real correctness fix, not just typing.

Verified locally against main: ruff clean, mypy (--follow-imports=skip) no issues, full suite 853 passed / 3 skipped (optional extras). Merged — this closes #90. Welcome aboard as a Whoosh contributor! Happy to point you at the next good-first-issue whenever you'd like.

priya-sundaram-dev:
🎉 Your `whoosh.externalsort` type hints just shipped in **whoosh3 3.41.0** — now on PyPI (`pip install -U whoosh3`). Thanks again, @ShlokShar! https://github.com/priya-sundaram-dev/whoosh/releases/tag/v3.41.0

Issue #97, opened 2026-08-20
Add type hints to whoosh.reading.IndexReader (abstract base class)


priya-sundaram-dev:
Whoosh is on an ongoing journey to a fully type-annotated public API (see #3). Several modules — `whoosh.analysis`, `whoosh.idsets`, `whoosh.formats`, `whoosh.externalsort` (#90/#91), `whoosh.fields` (#93/#94), `whoosh.scoring`, `whoosh.highlight`, and now `whoosh.columns` (#95/#96) — are typed. A great next target is the **`IndexReader` abstract base class** in `src/whoosh/reading.py`, which defines the core reader interface the whole query layer talks to.

### Scope (kept small on purpose)
Annotate only the **`IndexReader` base class** (the class starting around line 172), i.e. the abstract interface methods. Most of them are one-liners that either `raise NotImplementedError` or return a simple property, so this is mechanical and low-risk. Representative methods:
- Context/dunder: `__enter__`, `__exit__`, `__contains__`, `__iter__`
- Accessors: `codec`, `segment`, `segments`, `storage`, `is_atomic`, `close`, `generation`
- Term iteration: `terms_from`, `field_terms`, `iter_from`, `iter_field`, `iter_prefix`, `all_terms`
- Docs/lengths: `all_doc_ids`, `iter_docs`, `field_length`, `min_field_length`, `max_field_length`, `doc_field_length`
- Postings/vectors: `iter_postings`, `vector`, `vector_as`
- Misc: `corrector`, `terms_within`, `most_frequent_terms`, `most_distinctive_terms`, `leaf_readers`, `supports_caches`, `has_column`, `column_reader`

The concrete readers — `SegmentReader`, `MultiReader`, `EmptyReader` (and `MultiCursor`) — are intentionally **out of scope** here; they can follow in a second PR. The `TermInfo` class is also out of scope.

### How to do it
1. Add `from __future__ import annotations` at the top of the module (if not already present).
2. Put any import-cycle-prone imports (e.g. `whoosh.fields.Schema`, `whoosh.codec` types, `whoosh.matching.Matcher`) under `if TYPE_CHECKING:` — the `__future__` import makes all annotations lazy, so there's no runtime cost.
3. Annotate parameters and return types. Good reference modules for our patterns: `whoosh/scoring.py`, `whoosh/columns.py` (just merged in #96), `whoosh/fields.py`.
[...]

PR #100, opened 2026-08-21
perf: skip zlib for postings blocks below the measured crossover

### LINKED ISSUE #99: Skip zlib compression on tiny postings blocks (whoosh3 codec)
### Background

While profiling `IndexWriter.commit()` on a high-vocabulary corpus, `zlib.compress` in the `whoosh3` codec's `W3PostingsWriter._write_block` shows up as a real slice of index-build time. Most of that vocabulary is Zipfian: a large fraction of terms occur only a handful of times, so their postings live in a single small block.

For those tiny blocks, zlib often does no useful work — the smallest real block pickle is ~56 bytes, and compressing ~50–90 bytes of fairly high-entropy data tends to *expand* it (zlib adds a small header/checksum) while still costing CPU. There is already a "don't bother compressing" guard in `_write_block`, but its threshold is `< 20` bytes, which the payload never reaches, so it never fires. (A dead-store bug that hid this was just fixed in b211c81f — the guard now works, but the threshold is still conservative.)

### The task

Investigate raising the small-block compression-skip threshold so single-posting / very small blocks are stored uncompressed, and measure the effect:

1. Instrument `_write_block` to record, per block, the pre-compression pickle length and the post-compression length.
2. On a representative corpus (`benchmark/reuters.py` / `benchmark/regression.py` are good starting points), find the crossover size below which compression stops shrinking the payload.
3. Set the threshold to that crossover (as a named constant), and confirm via `benchmark/regression.py --compare` that index-build time improves (or is neutral) and that **on-disk index size does not regress**.
4. Add a small regression test asserting a round-trip on an index containing single-posting terms.

### Why it's a good first issue

- Self-contained: one method (`src/whoosh/codec/whoosh3.py::_write_block`) plus a test.
- The read path already honours `comp == 0` (`_read_data` only decompresses when the stored compression flag is truthy), so uncompressed small blocks are already format-legal — no format-version bump needed.
- It comes with a clear, measurable success criterion (the benchmark harness).

Happy to review a draft and pair on the benchmark numbers. Please comment to claim it.

tltaylor1:
## Summary

Raises the small-block compression-skip threshold in the whoosh3 codec from 20 to a measured crossover of 80 bytes, as a named constant (COMPRESSION_MIN_SIZE). Instrumenting _write_block on the reuters benchmark corpus showed no pickled block under 80 bytes ever compressed smaller, and that range is dominated by single-posting blocks of rare terms (4194 of 8989 blocks in the sample), each stored expanded by zlib's header while costing CPU. From 80 bytes up, compression always helped.

Measured effect, best of three builds on the reuters corpus: build time 0.462s against 0.499s, and the on-disk index shrinks by 4185 bytes, so size does not regress. Happy to rerun numbers with the maintainer's preferred harness settings.

## Related issues

Fixes #99

## Checklist

- [x] Tests pass locally (`pytest tests/`): 854 passed, 3 skipped
- [x] Added/updated tests: a round-trip test over single-posting terms, plus a zlib tripwire that pins the guard and fails against the old threshold
- [ ] Updated docs / CHANGELOG if user-facing: internal codec behavior, no user-facing surface changed
- [x] I have read the CONTRIBUTING guide

Note on ruff: the new test uses function-level imports to match the established style of tests/test_codecs.py; ruff flags these the same way it flags the file's pre-existing tests.

### DIFF
--- a/src/whoosh/codec/whoosh3.py
+++ b/src/whoosh/codec/whoosh3.py
@@ -65,6 +65,10 @@ except ImportError:
 # codec/version
 WHOOSH3_HEADER_MAGIC = b"W3Bl"
 
+# Below this pickled-block size, zlib reliably expands the payload
+# (measured crossover on the reuters benchmark corpus; see issue #99).
+COMPRESSION_MIN_SIZE = 80
+
 # Column type to store field length info
 LENGTHS_COLUMN = columns.NumericColumn("B", default=0)
 # Column type to store pointers to vector posting lists
@@ -816,16 +820,16 @@ class W3PostingsWriter(base.PostingsWriter):
         # Pickle the tuple
         databytes = dumps(data, 2)
         # Compress the pickle (if self._compression > 0), but skip compression
-        # for a payload so small that zlib's header would only add overhead
-        # instead of shrinking it. NOTE: the previous code set ``comp = 0`` here
-        # and then immediately overwrote it with ``comp = self._compression``,
-        # so this tiny-block skip never actually took effect. Ordered correctly
-        # now. (The threshold is conservative — the smallest real block pickle
-        # is ~56 bytes — so this is behaviourally a no-op today; raising it to
-        # skip compression on single-posting blocks of rare terms is tracked as
-        # a follow-up optimization.)
+        # for payloads below COMPRESSION_MIN_SIZE, where zlib's header and
+        # checksum expand the data instead of shrinking it while still
+        # costing CPU. Measured on the reuters benchmark corpus: no block
+        # under 80 bytes ever compressed smaller, and the sub-80 range is
+        # dominated by single-posting blocks of rare terms (roughly half of
+        # all blocks in a Zipfian vocabulary); from 80 bytes up, compression
+        # always helped. The read path honours the stored compression flag,
+        # so uncompressed blocks are format-legal.
         comp = self._compression
-        if len(databytes) < 20:
+        if len(databytes) < COMPRESSION_MIN_SIZE:
             comp = 0
         if comp:
             databytes = zlib.compress(databytes, comp)
--- a/tests/test_codecs.py
+++ b/tests/test_codecs.py
@@ -827,3 +827,67 @@ def test_whoosh2_legacy_codec_importable():
     ):
         assert hasattr(w2, name), name
     assert issubclass(w2.NoGraphError, Exception)
+
+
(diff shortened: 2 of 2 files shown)


tltaylor1:
Ran `benchmark/regression.py --compare` against a fresh baseline from main, per the notes on #99: no metric beyond tolerance. `build_index_ms` -0.9%, queries within noise, `incremental_add_ms` +12%, which is 0.35 ms of jitter on a 3 ms operation.

priya-sundaram-dev:
Thank you for this, @tltaylor1 — this is exactly the kind of careful, measured contribution I love to see. A few things I especially appreciate:

- **Measured, not guessed.** Instrumenting `_write_block` on the reuters corpus to find the real 80-byte crossover (rather than picking a round number) is the right way to do this, and matching it to the single-posting-block/Zipfian argument makes the *why* obvious to the next reader.
- **The tripwire test.** Swapping in a `TripwireZlib` that asserts no sub-threshold payload is ever handed to `zlib` pins the guard against silent regressions far better than a plain round-trip would. Nice touch pairing it with the round-trip so we cover both correctness and the optimization itself.
- **Format-legality reasoning is correct.** The per-block `comp` flag is written into the block header and the read path honours it (`if self._compression: b = zlib.decompress(b)`), so uncompressed tiny blocks are fully format-legal — good call spelling that out in the comment.

On the ruff note: you're right that the function-level imports match the existing style of `test_codecs.py`, so no need to change them — I'll keep the file consistent.

I've approved the CI workflow run (first-time-contributor gate). Once it goes green I'll merge. Really glad to have you working on this. 🙏

priya-sundaram-dev:
Merged — thank you, @tltaylor1! 🎉 This is your work landing in `main`, and it shipped clean across the full matrix (py3.9–3.14, Windows/macOS, ruff, mypy). It'll go out in the next whoosh3 release. If you enjoyed this one, `#97` (partial `IndexReader` method) is open and in a similar spirit — no pressure at all, but you'd be very welcome back.

priya-sundaram-dev:
🚀 Shipped in [**whoosh3 3.42.0**](https://pypi.org/project/whoosh3/3.42.0/) (`pip install -U whoosh3`), out on PyPI now. Thanks again @tltaylor1 — the measured crossover analysis and round-trip test were exactly the kind of rigor this codec work needs. You are credited in the [changelog](https://github.com/priya-sundaram-dev/whoosh/blob/main/CHANGELOG.md) and the release notes. Hope to see you back for another one!

PR #104, opened 2026-08-23
Do not ignore the results of the ruff linter

cclauss:
## Summary

Do not ignore the results of the ruff linter; it is finding real problems that are being ignored.

## Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,12 +18,12 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest]
-        python-version: ["3.9", "3.10", "3.11", "3.12", "3.13", "3.14"]
+        python-version: ["3.9", "3.10", "3.11", "3.12", "3.13", "3.14", "3.14t"]
         include:
           - os: windows-latest
-            python-version: "3.12"
+            python-version: "3.x"
           - os: macos-latest
-            python-version: "3.12"
+            python-version: "3.x"
     steps:
       - uses: actions/checkout@v7
       - name: Set up Python ${{ matrix.python-version }}
@@ -46,7 +46,7 @@ jobs:
       - uses: actions/checkout@v7
       - uses: actions/setup-python@v7
         with:
-          python-version: "3.13"
+          python-version: "3.x"
           allow-prereleases: true
       - name: Install
         run: |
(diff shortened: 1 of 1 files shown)


priya-sundaram-dev:
Thanks @cclauss — you're right, and this is a good catch. The `|| true` meant the lint job was decorative: `ruff check .` was surfacing 457 findings that never failed the build.

I've pushed the cleanup to `main` (06df459) so the gate can pass honestly rather than by being ignored:

- **Fixed the ~80 genuine nits** ruff was right about — double quotes, sorted imports, dropped `u""` prefixes, `Optional[x]` → `x | None` (only in modules that already `from __future__ import annotations`, so it stays 3.9-safe), `Tuple` → `tuple`, dict `.items()` iteration, and a few comprehension/f-string cleanups.
- **Documented the intentional suppressions** for the patterns Whoosh relies on by design, instead of leaving them to silently pass: function-local lazy imports (`PLC0415`, used throughout to break import cycles and defer optional deps), naive datetimes in the `DATETIME` field (`DTZ`), and deliberate blind-except guards around optional integrations (`BLE001`).

I've merged `main` into this branch, so once CI re-runs your `|| true` removal should be green on top of the now-clean tree. Keeping the matrix additions here too (3.14t + `3.x` on Windows/macOS) — nice touch. Will merge once the checks come back green. Really appreciate the drive-by review. 🙏

priya-sundaram-dev:
Merged — thank you @cclauss! The lint gate now actually gates, and the free-threaded 3.14t + `3.x` matrix rows are a nice forward-looking addition. 🎉

cclauss:
Cool...  Python 3.9 is no longer supported, so that it can be dropped.
* https://devguide.python.org/versions/

Python 3.15 and 3.15t release candidates can be added to the testing as recommended at:
* https://www.python.org/downloads/release/python-3150rc1/

priya-sundaram-dev:
Done on 3.15 — added `3.15` and `3.15t` to the test matrix in 98448cc. `allow-prereleases` was already on, so setup-python resolves the RCs; I marked just those two rows `continue-on-error` so an upstream RC regression gives us early-warning signal without turning `main` red. They'll flip to blocking once 3.15.0 final ships.

On 3.9: you're right that it's past upstream EOL, and I'll drop it — but I'd like to do it as a clean, announced boundary rather than mid-stream. My plan is to keep the 3.9 row through the current 3.x line (the cost is basically zero while the tests stay green, and there's still a meaningful long-tail of pinned-3.9 deployments installing `whoosh3`), then raise the floor to 3.10 in the next minor and note it in the changelog + roadmap. If you feel strongly it should go sooner, I'm glad to be persuaded — happy to have that on the record in an issue.

Thanks again for the reviews here; the honest lint gate is a real improvement.

PR #105, opened 2026-08-23
Run pre-commit tests in CI

cclauss:
## Summary

Add the following command to our CI tests.
% `pre-commit run --all-files`

Even if a contributor does not have `pre-commit` installed locally, we can be sure that all pre-commit tests pass before reviewing their pull requests.

A more complete solution would be to add this repository to the https://pre-commit.ci service, which is free for open source projects like this one.  `pre-commit.ci` would automate the creation of pull requests like:
* #102 

Also accelerated and simplified the execution of Python command-line tools (like ruff and pre-commit) with `uv`.
* https://docs.astral.sh/uv/
* https://docs.astral.sh/uv/setup-uv/
* https://docs.astral.sh/uv/guides/tools/

## Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -67,12 +67,16 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v7
-      - uses: actions/setup-python@v7
-        with:
-          python-version: "3.x"
-      - run: python -m pip install ruff
-      - name: ruff check
-        run: ruff check .
+      - uses: astral-sh/setup-uv@20cfd1bf945f4377ade1205e4dbc17946fc9a30d # v10.0.1
+      - run: uvx ruff check
+
+  pre-commit:
+    name: Pre-commit
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v7
+      - uses: astral-sh/setup-uv@20cfd1bf945f4377ade1205e4dbc17946fc9a30d # v10.0.1
+      - run: uvx pre-commit run --all-files
 
   types:
     name: Types (mypy smoke)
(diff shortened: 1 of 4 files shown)


priya-sundaram-dev:
Thanks @cclauss — wiring pre-commit into CI is a good call, and moving the tooling to `uv`/`uvx` is a nice speedup.

The reason the new **Pre-commit** job was red here is that the repo's own `.pre-commit-config.yaml` had never actually been run over the tree. I've just landed that on `main`:

- **`ruff-format` project-wide** (70 files) + normalized a few missing trailing newlines — no semantic changes, full suite still green (865 passed, 2 skipped). 0d8cd9f9
- **Dropped the `flynt` hook.** It auto-rewrites every `%`/`str.format()` to f-strings across a legacy codebase, which is an opinionated mass rewrite I'd rather not take blind — `ruff` (lint + format) already covers our style. Happy to reconsider if you feel strongly; open to discussing in an issue.

Could you rebase this PR onto the updated `main`? The pre-commit job should then run against a clean base. If `ruff` 0.16.4 (pinned in the hook) surfaces any residual format drift vs. what I ran locally, I'll apply it. Once it's green I'll merge. Really appreciate you pushing on CI quality here.

priya-sundaram-dev:
Merged — thank you @cclauss! 🙏

Nice touch folding flynt into ruff's native **FLY** rule (with `FLY002` ignored) instead of the standalone hook — that keeps the whole formatter/linter toolchain in one place, and it's reproducible with plain `ruff` locally. Also grabbed the `TRIO`→`ASYNC1` selector cleanup for free.

Pre-commit is now enforced in CI on every PR, and moving to `uv`/`uvx` shaved the setup time. Appreciate you driving the CI-quality work across #104 and #105.

PR #108, opened 2026-08-23
Fix ruff rules F811 and F841

cclauss:
## Summary

Fix:
* % ` ruff rule F811 ` # https://docs.astral.sh/ruff/rules/redefined-while-unused/
* % ` ruff rule F841 ` # https://docs.astral.sh/ruff/rules/unused-variable/

## Checklist

### DIFF
--- a/benchmark/marc21.py
+++ b/benchmark/marc21.py
@@ -3,7 +3,7 @@ import logging
 import os.path
 import re
 
-from whoosh import analysis, fields, index, qparser, query, scoring
+from whoosh import analysis, fields, index, qparser, scoring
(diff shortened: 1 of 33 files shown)


priya-sundaram-dev:
Merged — thanks @cclauss! Validated locally: `ruff check --select F811,F841` clean, full format check clean, and the whole suite (867 tests) green. F811/F841 are now enforced in CI rather than ignored. Confirmed the duplicate `Composable` import in `analysis/__init__.py` was the F811-flagged shadow — the canonical import (and the `whoosh.analysis.Composable` public name) is retained. Appreciated.

PR #110, opened 2026-08-23
Do not skip pytests for langchain, llamaindex, and mcp

cclauss:
## Summary

Fix our pytest warnings so we do not skip tests.
```
=========================== short test summary info ============================
SKIPPED [1] tests/test_langchain.py:71: could not import 'langchain_core': No module named 'langchain_core'
SKIPPED [1] tests/test_llamaindex.py:44: could not import 'llama_index.core': No module named 'llama_index'
SKIPPED [1] tests/test_mcp.py:129: could not import 'mcp': No module named 'mcp'
866 passed, 3 skipped in 24.83s
```

## Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,7 +26,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest]
-        python-version: ["3.9", "3.10", "3.11", "3.12", "3.13", "3.14", "3.14t", "3.15", "3.15t"]
+        python-version: ["3.9", "3.10", "3.11", "3.12", "3.13", "3.14", "3.15"]
         include:
           - os: windows-latest
             python-version: "3.x"
@@ -44,7 +44,7 @@ jobs:
       - name: Install
         run: |
           python -m pip install --upgrade pip
-          python -m pip install -e .
+          python -m pip install -e ".[langchain,llamaindex,mcp]"
           python -m pip install pytest
       - name: Run tests
         run: python -m pytest tests/ -q
@@ -63,7 +63,7 @@ jobs:
       - name: Install
         run: |
           python -m pip install --upgrade pip
-          python -m pip install -e .
+          python -m pip install -e ".[langchain,llamaindex]"
           python -m pip install pytest
       - name: Run tests with DeprecationWarning promoted to error
         # Guards the "runs clean on modern Python" promise: if a future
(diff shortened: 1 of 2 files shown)


priya-sundaram-dev:
Thanks @cclauss — I'm fully on board with the goal here: unskipping the langchain / llamaindex / mcp integration tests turns three silently-skipped tests into real coverage, and the AI-integration examples are exactly the part of Whoosh I most want CI to guard. Two things are keeping it red, one trivial and one substantive:

**1. `pyproject-fmt` (trivial)** — it wants the markers with no spaces around `>`:
```toml
optional-dependencies.llamaindex = [ "llama-index-core>=0.10; python_version>'3.9'" ]
optional-dependencies.mcp = [ "mcp>=1,<2; python_version>'3.9'" ]
```

**2. Free-threaded rows fail on the extras (substantive).** Installing `.[langchain,llamaindex,mcp]` pulls in `orjson` transitively, and orjson currently ships no free-threaded wheel, so on `3.14t` / `3.15t` pip falls back to building it from source and maturin/cargo fails:
```
💥 maturin failed  —  ERROR: Failed building wheel for orjson
error: failed-wheel-build-for-install
```
`3.14t` is a blocking row, so this reds the whole run (`3.15t` is continue-on-error). Env markers can't express “not free-threaded,” so the cleanest fix is to install the extras everywhere except the `t` rows, e.g. split the Install step:
```yaml
      - name: Install (with AI integration extras)
        if: ${{ !endsWith(matrix.python-version, 't') }}
        run: |
          python -m pip install --upgrade pip
          python -m pip install -e ".[langchain,llamaindex,mcp]"
          python -m pip install pytest
      - name: Install (core only — free-threaded, no AI-extra wheels yet)
        if: ${{ endsWith(matrix.python-version, 't') }}
        run: |
          python -m pip install --upgrade pip
          python -m pip install -e .
          python -m pip install pytest
```
That keeps the free-threaded rows testing Whoosh itself while the integration tests run on every GIL build (incl. Windows/macOS). Happy to push these two tweaks onto the branch myself if you'd rather I take it from here — just say the word.

Separately, the `mcp>=1,<2` pin is a good catch and a real signal: our `examples/mcp_server.py` needs updating for the mcp 2.x API. I'll open a tracking issue for that so the upper bound has a clear path off.

priya-sundaram-dev:
Merged — thanks! Running the langchain/llamaindex/mcp tests for real instead of skipping them is a nice reliability win. I added one follow-up commit (119a677d): I restored the free-threaded `3.14t`/`3.15t` rows using a per-row `extras` matrix variable that installs base-only on free-threaded builds. Those integration deps have no free-threaded wheels yet, but the pure-Python core needs to keep its free-threaded coverage — those rows have caught real threading races (e.g. the AsyncWriter join). Stable rows still install `.[langchain,llamaindex,mcp]` and exercise the integrations. 🙏

PR #112, opened 2026-08-23
Support MCP Python SDK 2.x in whoosh.mcp

### LINKED ISSUE #111: Support MCP Python SDK 2.x in whoosh.mcp (currently pinned to mcp>=1,<2)
## Context

PR #110 pins the `mcp` optional dependency to `mcp>=1,<2` so CI can run `tests/test_mcp.py` against a known-good SDK. That upper bound is load-bearing, not cosmetic: **MCP Python SDK 2.0 removed the `mcp.server.fastmcp` module.** `src/whoosh/mcp.py` imports:

```python
from mcp.server.fastmcp import FastMCP   # build_mcp_server()
```

Under SDK 2.x this raises `ModuleNotFoundError: No module named 'mcp.server.fastmcp'`, so `whoosh-mcp` / `build_mcp_server()` would fail to start for anyone who installs the current SDK.

## Goal

Migrate `whoosh.mcp` to the 2.x server API while keeping the dependency-free `SearchCore` untouched, then drop the `<2` upper bound.

## Tasks

- [ ] Determine the 2.x import path / entry point that replaces `mcp.server.fastmcp.FastMCP` and the `@server.tool()` decorator surface used in `build_mcp_server()`.
- [ ] Update `build_mcp_server()` (and the `whoosh-mcp` console script `main()`) accordingly; keep the clean "SDK not installed" error path.
- [ ] Keep `examples/mcp_server.py` in sync (it wraps `whoosh.mcp`).
- [ ] Extend `tests/test_mcp.py` so the server-construction path is exercised when the SDK is present (it currently only asserts a clean failure when absent).
- [ ] Relax the pin to `mcp>=<new-min>` and remove the `# TODO: Fix code to support MCP 2.x` note.

## Notes

`SearchCore` has no MCP dependency and is unaffected — this is purely the server wrapper. Good first issue for someone who's used the MCP SDK; happy to guide. Tracking the pin added in #110.

mayuriphad:
Closes #111

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -39,7 +39,7 @@ dynamic = [ "version" ]
 optional-dependencies.dev = [ "mypy", "pre-commit", "pytest>=7", "pytest-cov", "ruff" ]
 optional-dependencies.langchain = [ "langchain-core>=0.3" ]
 optional-dependencies.llamaindex = [ "llama-index-core>=0.10; python_version>'3.9'" ]
-optional-dependencies.mcp = [ "mcp>=2.0.0; python_version>'3.9'" ]
+optional-dependencies.mcp = [ "mcp>=2; python_version>'3.9'" ]
 urls.Changelog = "https://github.com/priya-sundaram-dev/whoosh/blob/main/CHANGELOG.md"
 urls.Documentation = "https://priya-sundaram-dev.github.io/whoosh/"
 urls.Homepage = "https://github.com/priya-sundaram-dev/whoosh"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -39,8 +39,7 @@ dynamic = [ "version" ]
 optional-dependencies.dev = [ "mypy", "pre-commit", "pytest>=7", "pytest-cov", "ruff" ]
 optional-dependencies.langchain = [ "langchain-core>=0.3" ]
 optional-dependencies.llamaindex = [ "llama-index-core>=0.10; python_version>'3.9'" ]
-# TODO: Fix code to support MCP 2.x and remove the upper bound.
-optional-dependencies.mcp = [ "mcp>=1,<2; python_version>'3.9'" ]
+optional-dependencies.mcp = [ "mcp>=2.0.0; python_version>'3.9'" ]
 urls.Changelog = "https://github.com/priya-sundaram-dev/whoosh/blob/main/CHANGELOG.md"
 urls.Documentation = "https://priya-sundaram-dev.github.io/whoosh/"
 urls.Homepage = "https://github.com/priya-sundaram-dev/whoosh"
--- a/src/whoosh/mcp.py
+++ b/src/whoosh/mcp.py
@@ -54,7 +54,7 @@ from whoosh.qparser import MultifieldParser
 if TYPE_CHECKING:  # pragma: no cover - typing only, avoids importing the SDK
     from collections.abc import Sequence
 
-    from mcp.server.fastmcp import FastMCP
+    from mcp.server import MCPServer
 
 __all__ = ["SearchCore", "build_mcp_server", "main", "DEFAULT_EXTS", "SAMPLE_DOCS"]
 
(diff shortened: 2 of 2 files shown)


priya-sundaram-dev:
Thank you for this, @mayuriphad — and welcome! 🎉 This is a genuinely useful contribution and your core insight is spot on: **MCP SDK 2.0 renamed `FastMCP` to `MCPServer` (now at `mcp.server.MCPServer`), while keeping the `@server.tool()` / `server.run()` surface unchanged.** That's exactly the breakage #111 is about, and you found the right replacement API.

There's one thing I'd like to resolve before we merge: the change as written **drops every user still on the official MCP SDK 1.x**. Swapping `from mcp.server.fastmcp import FastMCP` → `from mcp.server import MCPServer` and hard-pinning `mcp>=2.0.0` means anyone who currently has `whoosh3[mcp]` installed against 1.x gets an `ImportError` on upgrade (and 2.x also requires Python ≥ 3.10, whereas the extra currently supports 3.9 via the env marker). Since the MCP ecosystem is mid-migration, I'd rather support **both** rather than force everyone to 2.x at once.

The good news is the two APIs are drop-in compatible for what `whoosh.mcp` uses, so a tiny loader handles every case with zero user breakage:

```python
def _load_fastmcp():
    # Official MCP SDK 1.x
    try:
        from mcp.server.fastmcp import FastMCP
        return FastMCP
    except ModuleNotFoundError:
        pass
    # Official MCP SDK 2.x (FastMCP renamed to MCPServer) -- your fix
    try:
        from mcp.server import MCPServer
        return MCPServer
    except (ModuleNotFoundError, ImportError):
        pass
    # Standalone `fastmcp` 2.x package
    try:
        from fastmcp import FastMCP
        return FastMCP
    except ModuleNotFoundError as exc:
        raise ModuleNotFoundError(
            'The MCP server requires an MCP SDK. Install:  pip install "whoosh3[mcp]"'
        ) from exc
```

`build_mcp_server()` then just calls `Server = _load_fastmcp()` instead of importing directly, and the `mcp` extra can stay `mcp>=1` (no upper bound needed anymore, since the code handles both majors).

Would you like to update this PR with that approach? Happy to walk through anything, and I can push the loader to your branch or co-author it if you'd prefer I take it from here — either way the credit for finding the 2.x API is yours. Thanks again for jumping on #111 so quickly. 🙏

pyproject.toml
```diff
@@ -39,8 +39,7 @@ dynamic = [ "version" ]
 optional-dependencies.dev = [ "mypy", "pre-commit", "pytest>=7", "pytest-cov", "ruff" ]
 optional-dependencies.langchain = [ "langchain-core>=0.3" ]
 optional-dependencies.llamaindex = [ "llama-index-core>=0.10; python_version>'3.9'" ]
-# TODO: Fix code to support MCP 2.x and remove the upper bound.
-optional-dependencies.mcp = [ "mcp>=1,<2; python_version>'3.9'" ]
+optional-dependencies.mcp = [ "mcp>=2.0.0; python_version>'3.9'" ]
```
cclauss:
```suggestion
optional-dependencies.mcp = [ "mcp>=2; python_version>'3.9'" ]
```
To placate pyproject-fmt in pre-commit.
https://github.com/priya-sundaram-dev/whoosh/blob/70000b40f29d5d81ce07f1484ffc4df6ca97a4d3/.pre-commit-config.yaml#L46

Running `pre-commit install` locally will auto-fix this stuff.

priya-sundaram-dev:
Update — and a course-correction from me. 🙏

I went and **verified your change against the released SDK**: MCP `2.0.0` is now on PyPI, the `FastMCP → MCPServer` rename (modelcontextprotocol/python-sdk#1732) shipped in it, and the official quickstart is exactly `from mcp.server import MCPServer` / `MCPServer("…")` / `@server.tool()` — precisely what you wrote. Because this PR branch already carries the CI change that installs `.[…,mcp]` on the stable rows, `test_build_mcp_server_exposes_search_and_fetch_when_sdk_present` **actually ran against mcp 2.0.0 and passed** here — so this is verified end-to-end, not just importing cleanly. Nice work.

On my earlier ask for a 1.x/2.x compatibility loader: **I've talked myself out of it — let's go with your clean 2.x-only switch.** My reasoning changed once I looked at how it'd be tested: if we left the pin at `mcp>=1`, CI would install the latest (2.x) and the 1.x code path would never actually execute — i.e. we'd be shipping an untested branch to "support" a legacy SDK on a brand-new optional feature. A single, CI-exercised requirement on the current SDK is the more honest and maintainable choice. So please disregard the loader suggestion above; your approach is the one I want to merge.

Only remaining thing was the `pyproject-fmt` pre-commit nit that @cclauss caught (it normalizes `mcp>=2.0.0` → `mcp>=2`). To save you a round-trip I pushed that one-line normalization to your branch and co-authored it to Christian. Once CI comes back green I'll merge. Thank you both — @mayuriphad for finding and wiring up the 2.x API, @cclauss for the fmt catch. This closes #111. 🎉

PR #116, opened 2026-08-24
types(matching): annotate the Matcher base-class interface

### LINKED ISSUE #115: Add type hints to whoosh.matching.mcore.Matcher (abstract base class)
Whoosh is on an ongoing journey to a fully type-annotated public API. Several core modules are already typed — `whoosh.analysis`, `whoosh.fields`, `whoosh.scoring`, `whoosh.highlight`, `whoosh.columns`, and most recently the `IndexReader` abstract base class in `whoosh.reading` (#97). A great next target is the sibling of that reader interface on the query side: the **`Matcher` abstract base class** in `src/whoosh/matching/mcore.py`.

`Matcher` defines the interface every posting iterator implements (term matchers, boolean combinators, wrappers), so getting its signatures right makes the whole `whoosh.matching` subpackage easier to annotate later.

Annotate **only the `Matcher` base class** (starts at line 74), i.e. the abstract interface methods. Most are one-liners that either `raise NotImplementedError` or return a simple value, so this is mechanical and low-risk. Representative methods:
- State/iteration: `is_active`, `reset`, `id`, `next`, `skip_to`, `copy`, `depth`
- Terms: `term`, `term_matchers`, `matching_terms`, `is_leaf`, `children`, `replace`
- Quality/scoring: `supports_block_quality`, `max_quality`, `block_quality`, `skip_to_quality`, `weight`, `score`
- Values/spans: `all_ids`, `all_items`, `items_as`, `value`, `supports`, `value_as`, `spans`
- Rich-comparison dunders: `__eq__`, `__lt__`, `__ne__`, `__gt__`, `__le__`, `__ge__`

The concrete subclasses in `matching/` (`NullMatcher`, `ListMatcher`, `WrappingMatcher`, etc.) are intentionally **out of scope** here — they can follow in a second PR.

### How to do it
1. Add `from __future__ import annotations` at the top of the module (it has none yet).
2. Put any import-cycle-prone imports (e.g. `whoosh.scoring`, span types) under `if TYPE_CHECKING:` — with the `__future__` import, annotations are lazy so there's no runtime cost.
3. Annotate parameters and return types. Good reference: `whoosh/reading.py` (the `IndexReader` ABC from #97), `whoosh/scoring.py`.
4. Common hints for this layer: `id`/`docnum` are `int`; `weight`/`score`/`quality` are `float`; `astype` is `str`; iterator methods return `Iterator[int]` / `Iterator[tuple[int, Any]]`; abstract stubs that `raise NotImplementedError` should still get a concrete return annotation (`Any` is fine where the value type is genuinely open, e.g. `value()`).

### Before you open the PR
```
python -m ruff check src/whoosh/matching/mcore.py
python -m mypy --follow-imports=skip src/whoosh/matching/mcore.py
python -m pytest tests/ -q
```

Georgefifth:
## Summary

Annotates **only the `Matcher` ABC** in `src/whoosh/matching/mcore.py` as scoped in the issue — 32 methods get parameter and return hints, following the `IndexReader` pattern from #97 (`from __future__ import annotations` at the top, cycle-prone `Span` import under `TYPE_CHECKING`). Annotations only; no runtime behavior changes. Concrete subclasses (`NullMatcher`, `ListMatcher`, wrappers, …) stay out of scope for a follow-up, as requested.

Representative choices: `id`/`docnum` as `int`, `weight`/`score`/`quality` as `float`, `term()` as `tuple[str, bytes] | None`, iterator methods as `Iterator[int]` / `Iterator[tuple[str, bytes]]`, `value()`/`value_as()` as `Any` since the encoded form is genuinely open, comparison dunders taking `object` per stubs conventions.

## Related issues

Fixes #115

## Checklist

- [x] Tests pass locally (`pytest tests/`) — 866 passed, 3 skipped (optional integrations not installed). Note: a first-ever run of the suite reports 2 demo-page failures because the tests themselves rewrite the version number in `demo/*.html`; a clean second run is green. This reproduces identically on untouched `main` and is unrelated to this change.
- [x] Added/updated tests for the change where it makes sense — annotations only, no behavior change to test
- [x] Updated docs / CHANGELOG if user-facing — not user-facing; skipped

Pre-PR verification as listed in the issue: `python -m ruff check src/whoosh/matching/mcore.py` clean; `python -m mypy --follow-imports=skip src/whoosh/matching/mcore.py` reports only one pre-existing `[abstract]` error at the `NullMatcherClass` singleton instantiation, which reproduces on untouched `main` (and touches a concrete subclass that is explicitly out of scope here).

### DIFF
--- a/src/whoosh/matching/mcore.py
+++ b/src/whoosh/matching/mcore.py
@@ -49,8 +49,16 @@ these optimizations, the matcher's :meth:`Matcher.supports_block_quality()`
 method will return ``True``.
 """
 
+from __future__ import annotations
+
 from abc import abstractmethod
 from itertools import repeat
+from typing import TYPE_CHECKING, Any
+
+if TYPE_CHECKING:
+    from collections.abc import Iterator
+
+    from whoosh.query.spans import Span
 
 # Exceptions
 
@@ -75,7 +83,7 @@ class Matcher:
     """Base class for all matchers."""
 
     @abstractmethod
-    def is_active(self):
+    def is_active(self) -> bool:
         """Returns True if this matcher is still "active", that is, it has not
         yet reached the end of the posting list.
         """
@@ -83,7 +91,7 @@ class Matcher:
         raise NotImplementedError
 
     @abstractmethod
-    def reset(self):
+    def reset(self) -> None:
         """Returns to the start of the posting list.
 
         Note that reset() may not do what you expect after you call
@@ -93,14 +101,14 @@ class Matcher:
 
         raise NotImplementedError
 
-    def term(self):
+    def term(self) -> tuple[str, bytes] | None:
         """Returns a ``("fieldname", "termtext")`` tuple for the term this
         matcher matches, or None if this matcher is not a term matcher.
         """
 
         return None
 
-    def term_matchers(self):
+    def term_matchers(self) -> Iterator[Matcher]:
         """Returns an iterator of term matchers in this tree."""
 
         if self.term() is not None:
@@ -109,7 +117,7 @@ class Matcher:
             for cm in self.children():
                 yield from cm.term_matchers()
 
-    def matching_terms(self, id=None):
+    def matching_terms(self, id: int | None = None) -> Iterator[tuple[str, bytes]]:
         """Returns an iterator of ``("fieldname", "termtext")`` tuples for the
         **currently matching** term matchers in this tree.
         """
@@ -130,17 +138,17 @@ class Matcher:
         else:
             yield t
 
-    def is_leaf(self):
+    def is_leaf(self) -> bool:
         return not bool(self.children())
 
-    def children(self):
+    def children(self) -> list[Matcher]:
         """Returns an (possibly empty) list of the submatchers of this
         matcher.
         """
 
         return []
 
-    def replace(self, minquality=0):
+    def replace(self, minquality: float = 0) -> Matcher:
         """Returns a possibly-simplified version of this matcher. For example,
         if one of the children of a UnionMatcher is no longer active, calling
         this method on the UnionMatcher will return the other child.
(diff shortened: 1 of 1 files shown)


priya-sundaram-dev (approved):
Thanks for the thorough, cleanly-scoped PR, @Georgefifth — and welcome as a first-time contributor! 🎉

This is exactly what #115 asked for: annotations only on the `Matcher` ABC, mirroring the #97 `IndexReader` pattern (`from __future__ import annotations` + the cycle-prone `Span` import guarded under `TYPE_CHECKING`). Your type choices are all sensible:

- `id`/`docnum` → `int`, `weight`/`score`/`quality`/`minquality` → `float` ✔
- `term()` → `tuple[str, bytes] | None`, iterators as `Iterator[int]` / `Iterator[tuple[str, bytes]]` ✔
- `value()`/`value_as()` → `Any` — correct, the encoded/decoded form is genuinely open ✔
- comparison dunders taking `object` per typeshed conventions ✔

I validated locally: `ruff check` clean, `test_matching.py` (27) + `test_spans.py` (21) green, import sanity OK. Your note about the `demo/*.html` first-run version-rewrite and the pre-existing `NullMatcherClass [abstract]` mypy error both reproduce on untouched `main` and are unrelated — appreciate you calling those out explicitly.

I approved the CI workflow (first-time-contributor gate). Once it goes green I'll merge. Great work.

priya-sundaram-dev:
Merged — thank you, @Georgefifth! 🙏 Your first contribution to Whoosh is in. The Matcher ABC is now fully annotated, matching the IndexReader work from #97. If you'd like a follow-up, the concrete subclasses (NullMatcher, ListMatcher, and the wrapper matchers in matching/) are the natural next step and I'm happy to open a scoped issue for it. Welcome aboard!

PR #118, opened 2026-08-24
chore: add type annotations to the Matcher base class

### LINKED ISSUE #115: Add type hints to whoosh.matching.mcore.Matcher (abstract base class)

Georgefifth:
Closes #115.

**Problem**
The `Matcher` base class in `src/whoosh/matching/mcore.py` is the most used
public type in the package, but none of its methods are annotated, so the
type-checking work done for the reader layer (gh#67) stops there.

**Solution**
Annotate every method of `Matcher` with the exact same pattern used by the
annotated `IndexReader` methods in `src/whoosh/reading.py`. Pure annotations,
no runtime behavior changes. Also extend the typing smoke fixture so the new
annotations are actually exercised by the CI mypy job.

**Result**
- `python -m mypy` (the CI gate, checks `tests/typing_smoke.py`)
- `python -m ruff check src/whoosh/matching/mcore.py`
- `python -m pytest tests/` (866 passed)

Note: `python -m mypy --follow-imports=skip src/whoosh/matching/mcore.py`
reports one pre-existing error about `NullMatcherClass` (present on main
before this change; that class is outside the scope of #115).

### DIFF
--- a/src/whoosh/matching/mcore.py
+++ b/src/whoosh/matching/mcore.py
@@ -49,8 +49,16 @@ these optimizations, the matcher's :meth:`Matcher.supports_block_quality()`
 method will return ``True``.
 """
 
+from __future__ import annotations
+
 from abc import abstractmethod
 from itertools import repeat
+from typing import TYPE_CHECKING, Any
+
+if TYPE_CHECKING:
+    from collections.abc import Iterator
+
+    from whoosh.query.spans import Span
 
 # Exceptions
 
@@ -75,7 +83,7 @@ class Matcher:
     """Base class for all matchers."""
 
     @abstractmethod
-    def is_active(self):
+    def is_active(self) -> bool:
         """Returns True if this matcher is still "active", that is, it has not
         yet reached the end of the posting list.
         """
@@ -83,7 +91,7 @@ class Matcher:
         raise NotImplementedError
 
     @abstractmethod
-    def reset(self):
+    def reset(self) -> None:
         """Returns to the start of the posting list.
 
         Note that reset() may not do what you expect after you call
@@ -93,14 +101,14 @@ class Matcher:
 
         raise NotImplementedError
 
-    def term(self):
+    def term(self) -> tuple[str, bytes] | None:
         """Returns a ``("fieldname", "termtext")`` tuple for the term this
         matcher matches, or None if this matcher is not a term matcher.
         """
 
         return None
 
-    def term_matchers(self):
+    def term_matchers(self) -> Iterator[Matcher]:
         """Returns an iterator of term matchers in this tree."""
 
         if self.term() is not None:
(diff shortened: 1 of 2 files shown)


Georgefifth:
Closing: #116 was merged to main while this PR was in flight and resolves the same issue (#115) with the same scope. No changes from mine are needed on top of #116. Sorry for the noise — happy to close out here.

PR #119, opened 2026-08-24
pre-commit: Add the extremely fast Python type checker ty

cclauss:
## Summary

Add an extremely fast Python type checker and language server, written in Rust.
* https://docs.astral.sh/ty

Pre-commit hooks should have subsecond run times, or when run locally, developers may lose context and focus.

`ty` is much faster and more capable than `mypy` and is backed by [Astral](https://astral.sh/), the creators of [uv](https://github.com/astral-sh/uv) and [Ruff](https://github.com/astral-sh/ruff).

Across many projects, I have found that `ty` is easier to deal with and provides more useful typing results than `mypy`.

## Checklist

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -43,6 +43,11 @@ repos:
         args: [ --fix ]
       - id: ruff-format
 
+  - repo: https://github.com/astral-sh/ty-pre-commit
+    rev: v0.0.74
+    hooks:
+      - id: ty
+
   - repo: https://github.com/abravalheri/validate-pyproject
     rev: '0.26'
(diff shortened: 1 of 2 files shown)


priya-sundaram-dev:
Thanks @cclauss! `ty` is scoped narrowly here (just `tests/typing_smoke.py`, same target as the existing mypy smoke) and pinned to `v0.0.74`, so it complements rather than replaces the mypy job while we watch `ty` mature toward 1.0. All 15 checks are green and the pre-commit run stays subsecond-fast. Appreciate you keeping the dev tooling sharp — merging. 🚀

Issue #117, opened 2026-08-24
Add type hints to the Index abstract base class (whoosh.index)


priya-sundaram-dev:
Whoosh is on an ongoing journey to a fully type-annotated public API. Several core interfaces are already done — the `IndexReader` abstract base class in `whoosh.reading` (#97) and, most recently, the `Matcher` abstract base class in `whoosh.matching.mcore` (#115). A great next target completes that trio of core interfaces: the **`Index` abstract base class** in `src/whoosh/index.py` (starts around line 214), plus the module-level convenience functions right above it.

`Index` is arguably the single most user-facing object in the whole library — it's what `open_dir()` / `create_in()` hand back — so getting its signatures right pays off everywhere downstream.

### Scope (kept small on purpose)
Two closely-related things in `src/whoosh/index.py`, and **nothing else**:

1. **Module-level functions** — they already have *parameter* annotations but are missing *return types* (and a couple, `version_in`/`version`/`clean_files`, have no annotations at all):
   - `create_in`, `open_dir`, `exists_in`, `exists`, `version_in`, `version`, `clean_files`
2. **The `Index` abstract base class methods** — currently unannotated one-liners, most of which either `raise NotImplementedError` or return a simple value:
   - Lifecycle: `close`, `add_field`, `remove_field`, `optimize`, `refresh`
   - Generation/freshness: `latest_generation`, `up_to_date`, `last_modified`, `is_empty`
   - Counts/lengths: `doc_count_all`, `doc_count`, `field_length`, `max_field_length`
   - Access: `searcher`, `reader`, `delete_by_term`, `delete_by_query`

The concrete `FileIndex` subclass lower in the file is intentionally **out of scope** here — it can follow in a second PR.

### How to do it
1. The module already has `from __future__ import annotations`, so annotations are lazy — put any import-cycle-prone imports (e.g. `whoosh.searching.Searcher`, `whoosh.reading.IndexReader`) under `if TYPE_CHECKING:` and reference them as strings/lazy names with no runtime cost.
2. Annotate return types. Good reference: `whoosh/reading.py` (the `IndexReader` ABC from #97).
[...]

PR #120, opened 2026-08-24
Type-check the entire codebase with ty, with lots of rules ignored

cclauss:
## Summary

Type-check the entire codebase with ty, with lots of rules ignored.  This puts guardrails in place so that `ty` can catch all typing rules that are not ignored.  Provide opportunities to unignore ty rules as we gradually implement Python typing.

Fixing typing rules can be complicated, so future typing pull requests should unignore one or two rules to simplify the code review process.

## Related issues
* #117

## Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -179,4 +179,27 @@ warn_redundant_casts = true
 warn_unused_ignores = true
 
 [tool.ty]
-src.include = [ "tests/typing_smoke.py" ]
+# TODO: Gradual typing: Revisit these ignores and remove them as the codebase is
+# incrementally typed (gh#3).  The instance comments below were calculated in the
+# past so they should be treated as a historical reference, not a current count.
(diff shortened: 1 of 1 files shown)


priya-sundaram-dev (approved):
Thanks @cclauss — this is exactly the right way to bring `ty` up on the whole codebase: turn it on everywhere, ignore the current rule failures explicitly, and leave a documented trail so we can un-ignore them one or two at a time in follow-up PRs. I verified `ty check` passes locally with this config and CI is green across all jobs (including the pre-commit `ty` hook).

One tiny note for a future pass (not blocking): the `gh#3` reference in the TODO doesn't map to an issue in this repo — I'll open a tracking issue for the gradual un-ignoring effort so subsequent PRs have a real anchor. Merging now. 🙌

cclauss:
* #117 might be the right substitute for gh#3

PR #128, opened 2026-08-24
.readthedocs.yaml: Upgrade Ubuntu, Node.js, and Python

cclauss:
## Summary

Example: https://docs.readthedocs.com/platform/stable/guides/reproducible-builds.html

## Checklist


priya-sundaram-dev:
Merged — thank you! The RTD `python: "3.7"` pin was long overdue for a bump; 3.14 + ubuntu-24.04 lines it up with the rest of the toolchain now that the floor is 3.10. 🙏

Issue #121, opened 2026-08-24
Tracking: gradually un-ignore `ty` rules as the codebase gets typed


priya-sundaram-dev:
As of #120, `ty` runs across the whole codebase in pre-commit/CI, but with a list of rule failures explicitly ignored in `[tool.ty]` (`pyproject.toml`). This issue tracks the gradual work of removing those ignores one or two at a time.

**How to help:** pick one ignored rule from the `[tool.ty].rules` block in `pyproject.toml`, remove its `ignore` line, run `ty check src/whoosh`, fix (or narrowly `# ty: ignore[rule]`) the reported cases, and open a small focused PR. Please keep each PR to one or two rules so review stays easy.

**Current ignore list** (dated snapshot, ty 0.0.74 — re-run with a rule flipped to `warn` for a fresh count):

- `missing-argument` — 12  ← smallest remaining, good starting point
- `not-subscriptable` — 15
- `call-non-callable` — 15 (was 20; #144, #183, #184)
- `invalid-assignment` — 21
- `too-many-positional-arguments` — 27
- `unsupported-operator` — 27
- `unresolved-import` — 29
- `invalid-argument-type` — 30
- `invalid-method-override` — 131
- `unresolved-attribute` — 347

**Already cleared** (no longer ignored): `invalid-ignore-comment` (#144), `empty-body`, `parameter-already-assigned`, `invalid-type-form`, `invalid-raise`, `invalid-yield`, `invalid-parameter-default`, and `no-matching-overload`. Two of the last were re-surfaced by the `whoosh.util.text` typing work in #175 and fixed there (analyzer `expression` params + `dateparse` `parse` signatures), keeping those rules green.

A companion, lower-friction path that also feeds this effort: annotate one `whoosh.util.*` / analysis module at a time (see the "good first issue" typing series, e.g. #173 → done in #175). Cleaner signatures shrink the counts above naturally.

Thanks @cclauss for setting up the guardrails.

priya-sundaram-dev:
✅ `invalid-yield` un-ignored in #122 (thanks @cclauss). The `pyproject.toml` ignore list is the source of truth for what remains — good next candidates are the other low-instance rules (`invalid-type-form` (2), `no-matching-overload` (2), `invalid-raise` (3)). Each is a self-contained, newcomer-friendly typing fix.

priya-sundaram-dev:
[...]

PR #135, opened 2026-08-24
typing(index): add type hints to Index ABC and module functions

### LINKED ISSUE #117: Add type hints to the Index abstract base class (whoosh.index)

SantiagoDaleffe:
## Summary

Added return type annotations to the Index ABC and module-level functions in whoosh/index.py. Imports within TYPE_CHECKING were also sorted.

If everything looks good, I'd love to take care of the FileIndex subclass in a follow-up PR!

## Related issues

Fixes #117

## Checklist

### DIFF
--- a/src/whoosh/index.py
+++ b/src/whoosh/index.py
@@ -46,6 +46,8 @@ from whoosh.system import _FLOAT_SIZE, _INT_SIZE, _LONG_SIZE
 if TYPE_CHECKING:
     from whoosh.fields import Schema
     from whoosh.filedb.filestore import Storage
+    from whoosh.reading import IndexReader
+    from whoosh.searching import Searcher
     from whoosh.writing import IndexWriter
 
 _DEF_INDEX_NAME = "MAIN"
@@ -89,7 +91,7 @@ class EmptyIndexError(IndexError):
 # Convenience functions
 
 
-def create_in(dirname: str, schema: Schema, indexname: str | None = None) -> FileIndex:
+def create_in(dirname: str, schema: Schema, indexname: str | None = None) -> Index:
     """Convenience function to create an index in a directory. Takes care of
     creating a FileStorage object for you.
 
@@ -116,7 +118,7 @@ def open_dir(
     indexname: str | None = None,
     readonly: bool = False,
     schema: Schema | None = None,
-) -> FileIndex:
+) -> Index:
     """Convenience function for opening an index in a directory. Takes care of
     creating a FileStorage object for you. dirname is the filename of the
     directory in containing the index. indexname is the name of the index to
@@ -166,7 +168,7 @@ def exists(storage: Storage, indexname: str | None = None) -> bool:
     return storage.index_exists(indexname)
 
 
-def version_in(dirname, indexname=None):
+def version_in(dirname, indexname=None) -> tuple[tuple[int, ...] | None, int]:
     """Returns a tuple of (release_version, format_version), where
     release_version is the release version number of the Whoosh code that
     created the index -- e.g. (0, 1, 24) -- and format_version is the version
@@ -193,7 +195,7 @@ def version_in(dirname, indexname=None):
     return version(storage, indexname=indexname)
 
 
-def version(storage, indexname=None):
+def version(storage, indexname=None) -> tuple[tuple[int, ...] | None, int]:
     """Returns a tuple of (release_version, format_version), where
     release_version is the release version number of the Whoosh code that
     created the index -- e.g. (0, 1, 24) -- and format_version is the version
@@ -220,8 +222,7 @@ def version(storage, indexname=None):
 
         ix = storage.open_index(indexname)
         return (ix.release, ix.version)
-    except IndexVersionError:
-        e = sys.exc_info()[1]
+    except IndexVersionError as e:
         return (None, e.version)
 
 
@@ -231,14 +232,14 @@ def version(storage, indexname=None):
 class Index:
     """Represents an indexed collection of documents."""
 
-    def close(self):
+    def close(self) -> None:
         """Closes any open resources held by the Index object itself. This may
         not close all resources being used everywhere, for example by a
         Searcher object.
         """
         pass
 
-    def add_field(self, fieldname, fieldspec):
+    def add_field(self, fieldname, fieldspec) -> None:
         """Adds a field to the index's schema.
 
         :param fieldname: the name of the field to add.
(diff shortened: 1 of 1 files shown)


priya-sundaram-dev (approved):
Thank you for this, @SantiagoDaleffe — and welcome as a first-time contributor! 🎉 This is exactly the kind of clean, well-scoped typing work #117 was asking for.

What I like:
- Return annotations on the whole `Index` ABC surface plus the module-level convenience/helper functions — consistent with the style we landed in #97 (fields) and #116 (matcher).
- Widening `create_in`/`open_dir` to return `Index` rather than the concrete `FileIndex` is the right call for a public convenience API: callers should program against the ABC, and it keeps the door open for other backends.
- Nice bonus cleanup of the `version()` handler (`except IndexVersionError as e:` instead of `sys.exc_info()`), and sorting the `TYPE_CHECKING` imports.

Validated locally: `ruff` clean, `ty check src/whoosh/index.py` → all checks passed, and `tests/test_indexing.py` (33) all pass. Annotations resolve as expected.

I've approved the CI run (first-time-contributor gate). Once it's green I'll merge. And yes — please do take the `FileIndex` subclass in a follow-up, that's a perfect next step! 🙏

priya-sundaram-dev:
Quick update after CI ran — 13 of 14 jobs are green, but the **Types (mypy smoke)** job caught one thing, and it's actually my fault for cheering on the return-type widening in my first pass. Let me walk through it because it's a genuinely interesting typing subtlety:

```
tests/typing_smoke.py:143: error: "Index" has no attribute "schema"  [attr-defined]
tests/typing_smoke.py:350: error: "Index" has no attribute "schema"  [attr-defined]
```

The `Types` job type-checks `tests/typing_smoke.py`, a fixture that exercises the public API exactly as a downstream user would. It does `ix = index.create_in(...)` and then `ix.schema`. `schema` lives on the concrete **`FileIndex`**, not on the `Index` ABC — and `create_in`/`open_dir` really do always return a `FileIndex` (they build a `FileStorage` for you). So widening their annotation to `Index` makes mypy lose the `.schema` (and `.storage`, reader-reuse, etc.) that real callers rely on — a small regression in the annotations we ship to users' editors.

So the fix is just to keep those two convenience functions returning the concrete type:

```diff
-def create_in(dirname: str, schema: Schema, indexname: str | None = None) -> Index:
+def create_in(dirname: str, schema: Schema, indexname: str | None = None) -> FileIndex:
...
     schema: Schema | None = None,
-) -> Index:
+) -> FileIndex:
```

**Everything else in the PR is exactly right and should stay** — the `Index` ABC method annotations, the `TYPE_CHECKING` imports, and the nice `except IndexVersionError as e:` cleanup. Only those two `-> Index` need to go back to `-> FileIndex`.

I verified locally: with just that revert, `python -m mypy` → "Success: no issues found" and the rest of CI stays green. Push it to the same branch and CI will re-run (I'll re-approve the first-timer gate), then I'll merge. Sorry for the back-and-forth — thanks for bearing with the type checker! 🙏

priya-sundaram-dev:
Perfect — that's exactly it. 🎯 Keeping `create_in`/`open_dir` at `-> FileIndex` while annotating the `Index` ABC surface with the abstract types is precisely the right call: callers of the factory functions get the concrete `.schema`/`.storage` attributes, and the ABC documents its own contract. I re-ran everything locally against your latest commit — `ruff` clean, `ty` clean, mypy smoke green, and `test_indexing.py` (33 tests) all pass. I've approved the CI run; once it comes back green I'll merge. Thanks for the careful iteration on the typing subtlety here — really nice first contribution. 🙌

priya-sundaram-dev:
Merged! 🎉 Thanks so much, @SantiagoDaleffe — this is exactly the kind of incremental, well-tested typing work that moves #117 forward, and you handled the `FileIndex` vs `Index` subtlety gracefully. The `Index` ABC now advertises a fully-annotated contract. Hope to see you around the issue tracker again — #121 tracks the follow-on typing work if you're looking for more. 🙏

PR #136, opened 2026-08-24
pre-commit: Add auto-walrus

cclauss:
## Summary

pre-commit: Add https://pypi.org/project/auto-walrus

## Checklist

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -31,8 +31,13 @@ repos:
       - id: sort-simple-yaml
       - id: trailing-whitespace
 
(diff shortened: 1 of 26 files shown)


priya-sundaram-dev (approved):
Thanks @cclauss! Pulled this locally against latest `main` — `ruff` clean, mypy smoke green, and the full suite (866 passed, 3 skipped) is happy. I read through the rewrites and they're all behavior-preserving: the `magic := postfile.read(4)` case still binds unconditionally, and every other conversion is a plain assign-then-test collapsed into the guard. Adding auto-walrus to pre-commit keeps this consistent going forward. LGTM. 🚀

PR #137, opened 2026-08-24
Enable many ruff rules

cclauss:
## Summary

% `ruff check --select=E714,B009,PIE808,RSE102,RUF019,RUF051,SIM118,SIM300 --fix`

% `ruff rule SIM118 ` # in-dict-keys is conservative, so 4 of those issues are still in the codebase.

Also manually fix `ruff rule PIE810 ` # multiple-starts-ends-with

## Checklist

### DIFF
--- a/examples/static_site_search.py
+++ b/examples/static_site_search.py
@@ -62,11 +62,7 @@ def build_index(directory, indexdir="site_index"):
     with ix.writer() as writer:
         for root, dirs, files in os.walk(directory):
             for file in files:
-                if (
-                    file.endswith(".md")
-                    or file.endswith(".rst")
-                    or file.endswith(".txt")
-                ):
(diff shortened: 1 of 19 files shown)


priya-sundaram-dev (approved):
Thanks @cclauss! Reviewed all 19 files locally — these are semantically-equivalent idiomatic cleanups (`range(0, n)`→`range(n)`, chained `.endswith`/`.startswith`→tuple form via PIE810, `not x is y`→`x is not y`, `getattr(o, "const")`→attr access, dropped redundant `.keys()`, `del`→`pop(..., None)`), plus a sensible expansion of the enabled ruff rule set (PD/PGH/PIE/RSE/T10). ruff clean and full test matrix green across 3.10–3.15. Merging.

PR #138, opened 2026-08-24
Ruff rules E for pycodestyle errors

cclauss:
## Summary

Fix https://docs.astral.sh/ruff/rules/#error-e

Before: 
% `ruff check --select=E --statistics`
```
493	E501	line-too-long
 42	E712	true-false-comparison
 28	E731	lambda-assignment
 11	E741	ambiguous-variable-name
  6	E722	bare-except
  4	E721	type-comparison
  3	E402	module-import-not-at-top-of-file
  2	E743	ambiguous-function-name
Found 589 errors.
No fixes available (70 hidden fixes can be enabled with the `--unsafe-fixes` option).
```

## Checklist

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -98,7 +98,7 @@ pygments_style = "sphinx"
 
 # -- Options for sphinx_rtd_theme -----------------------------------------
 # https://github.com/snide/sphinx_rtd_theme
-import sphinx_rtd_theme
+import sphinx_rtd_theme  # noqa: E402
 
 html_theme = "sphinx_rtd_theme"
 html_theme_path = [sphinx_rtd_theme.get_html_theme_path()]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -68,6 +68,7 @@ lint.select = [
   "C90",    # McCabe cyclomatic complexity
   "DJ",     # flake8-django
   "DTZ",    # flake8-datetimez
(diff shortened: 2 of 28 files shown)


priya-sundaram-dev:
Reviewed and merging — thank you, this is a great cleanup. 

What I checked, since a few of these `E` rules can quietly change behavior:
- **E712** (`== True`/`== False`) → truthiness: safe here, all the touched attributes (`startexcl`, `endexcl`, `constantscore`, `is_active()`, `is_leaf()`) are genuine booleans.
- **E721** (`type(x) == str` → `type(x) is str`): correct.
- **E722** bare-`except` and **E743** ambiguous names (`I`/`O` in `lang/lovins.py`, they mirror the published Lovins algorithm's rule names so keeping them + `# noqa` is the right call): behavior preserved via `# noqa`.
- **E731** lambda→`def`: all closures preserved their default-arg captures (e.g. `_fn=fieldname`), so no late-binding surprise.

One tiny note, no action needed: in `support/base85.py` the `l / 4 * 5` → `length // 4 * 5` line lives in `b85encode`, which is Python-2-era dead code (it does `word /= 85` and `struct.pack(... (l + 4) / 5 ...)`, both of which already blow up on Py3). Your `//` change makes it marginally less broken; I'll file a separate issue to either fix or delete that module rather than expand scope here.

14/14 checks green. Merging. 🎯

PR #140, opened 2026-08-24
ty: un-ignore empty-body and parameter-already-assigned rules

SantiagoDaleffe:
## Summary

This PR tackles two single-instance typing rules to help clear `pyproject.toml`:
1. **`empty-body`**: Removed the dead `parse_` stub in `src/whoosh/qparser/default.py`. A quick `git grep` confirmed it had no callers.
2. **`parameter-already-assigned`**: Fixed a double assignment of the `boost` parameter in `GroupNode.apply` (`src/whoosh/qparser/syntax.py`) by removing the residual positional `self.type` argument.

Both rules have been successfully un-ignored in `pyproject.toml`.

## Related issues

References #121

## Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -188,7 +188,6 @@ warn_unused_ignores = true
 # parameter-already-assigned) are the smallest starting points.  Re-run
 # `ty check src/whoosh` with a rule flipped to "warn" to get a fresh count.
 rules.call-non-callable = "ignore"  # 20 instances
-rules.empty-body = "ignore"  # 1 instance
 rules.invalid-argument-type = "ignore"  # 30 instances
 rules.invalid-assignment = "ignore"  # 21 instances
 rules.invalid-ignore-comment = "ignore"  # 1 instance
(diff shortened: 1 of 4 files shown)


priya-sundaram-dev (approved):
Excellent second contribution, @SantiagoDaleffe — and this one is more than a typing cleanup. 🎉

When I reviewed the `GroupNode.apply` change locally I realized the residual positional `self.type` you removed was an **actual latent bug**: `GroupNode`/`SyntaxNode` have no `type` attribute, so `self.type` bound to the `nodes=` parameter while the real node list slid into `boost=` (hence `ty`'s `parameter-already-assigned` on `boost`). Any call through `GroupNode.accept()` → `apply()` would have raised `AttributeError`. I exercised it locally after your fix and `AndGroup(...).accept(fn)` now round-trips correctly, preserving `boost` and the subnodes. Nice catch that the type checker surfaced.

The `empty-body` removal is also correct — `parse_` was a dead `pass` stub with no callers (`git grep` confirms), and you updated `test_typing.py` to drop it from the annotated-entry-point list. Verified locally:

- `ty check src/whoosh` — no `empty-body` / `parameter-already-assigned` diagnostics remain
- `ruff` clean, `mypy` smoke green
- full suite: **866 passed, 4 skipped**
- all 14 CI checks green

Two single-instance rules down, and a real bug fixed on the way. Merging now — thank you! 🙏

PR #142, opened 2026-08-25
Remove the dead byte-oriented base85 codec

### LINKED ISSUE #139: Dead code: byte-oriented b85encode/b85decode in support/base85.py are broken on Python 3
Surfaced while reviewing #138.

`whoosh/support/base85.py` exposes two families:

1. The **integer** codec `to_base85` / `from_base85` — used by the numeric field machinery and covered by tests. ✅
2. The **byte** codec `b85encode` / `b85decode` — Python-2-era code that cannot run on Python 3:
   - `b85encode` does `word /= 85` (float) and slices `out[0:olen]` where `olen` picked up float division; the recent `// ` tweak in #138 only partially patches this.
   - `b85decode` does `struct.pack(">" + "L" * ((l + 4) / 5), ...)` — `str * float` raises `TypeError`.

Neither byte function is imported anywhere in the codebase or exercised by the test suite (that's why CI stays green regardless), so they're effectively dead.

**Proposed fix:** delete `b85encode`/`b85decode` (and the `b85chars`/`b85dec` byte tables if unused elsewhere), keeping the integer codec. If we'd rather keep a working byte codec, replace them with a thin wrapper over the stdlib `base64.b85encode`/`b85decode`. Either way we should add a small regression test.

Good first issue for someone who wants a self-contained cleanup — happy to review. `good first issue`, `help wanted`.

chmm195:
Closes #139.

Took the delete option rather than the stdlib-wrapper option, for a reason worth stating explicitly.

### Why not wrap `base64.b85encode`

The two alphabets are not interchangeable:

```python
>>> from whoosh.support.base85 import b85chars
>>> list(b85chars) == sorted(b85chars)          # this module
True
>>> # stdlib base64.b85 alphabet: 0-9A-Za-z!#$%&()*+-;<=>?@^_`{|}~
>>> list(std) == sorted(std)
False
```

That ordering is the entire reason the custom alphabet exists — the comment in the file says as much — and `whoosh2.py:2250` uses it under a heading that reads *"Functions for converting sortable representations to and from text"*:

```python
def sortable_int_to_text(x, shift=0):
    text = chr(shift) + to_base85(x, False)
```

So a stdlib-backed `b85encode` would be a different codec wearing the same name, in a module whose one distinguishing feature is sortability. Deleting is the honest option; if a working byte codec is ever wanted, callers can reach for `base64` directly and be explicit about which alphabet they mean.

### Confirming it really is dead

Neither function is referenced anywhere outside its own definition:

```
$ grep -rn "b85encode\|b85decode" --include=*.py . | grep -v support/base85.py
(no output)
```

And both raise on every input type, so no caller could have been relying on them even accidentally:

```
b85encode(b"hello") -> TypeError: can't concat str to bytes
b85encode("hello")  -> TypeError: a bytes-like object is required, not 'str'
b85decode("abcde")  -> TypeError: can't multiply sequence by non-int of type 'float'
```

`b85encode` is broken on both branches, not just the `word /= 85` one #139 mentions — it also does `text += "\0" * (4 - r)` on what `struct.unpack` then requires to be bytes.

### What is kept

`b85chars`, `b85dec`, `to_base85`, `from_base85`. `import struct` is dropped (only the byte codec used it). The module docstring said "generic base85 encoding and decoding functions", which is no longer accurate, so it now says what the module provides and why the alphabet cannot be swapped — to save the next person from re-proposing the stdlib wrapper.

### Tests

New `tests/test_base85.py`, 7 cases: both roundtrips, fixed width (5 and 10 chars — the thing that makes it sortable), the alphabet being in ASCII order and 85 characters, and the order-preserving property both directly and through `sortable_int_to_text`.

There was no coverage of this module at all before, which is why CI stayed green over broken code.

- `pytest tests/` — 869 passed, 8 skipped (platform/optional-dep skips, unchanged)
- `ruff check` — clean; `ruff format --check` — already formatted

### DIFF
--- a/src/whoosh/support/base85.py
+++ b/src/whoosh/support/base85.py
@@ -1,13 +1,15 @@
 """
-This module contains generic base85 encoding and decoding functions. The
-whoosh.util.numeric module contains faster variants for encoding and
-decoding integers.
+This module contains base85 encoding and decoding functions for integers.
+The whoosh.util.numeric module contains faster variants.
+
+Unlike the standard ascii85/base85 alphabets, the character set here is in
+ASCII order, so encoded strings sort in the same order as the integers they
+encode. whoosh.codec.whoosh2 relies on that property, so the alphabet cannot
+be swapped for the one in the stdlib base64 module.
 
 Modified from:
 http://paste.lisp.org/display/72815
 """
-
-import struct
 
 # Instead of using the character set from the ascii85 algorithm, I put the
 # characters in order so that the encoded text sorts properly (my life would be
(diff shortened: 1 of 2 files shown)


priya-sundaram-dev:
Thank you so much for this, @chmm195 — and I owe you an honest, transparent explanation.

Two of you independently picked up #139 within about 15 minutes of each other. #141 (from @SAY-5) opened first and was complete end to end — same delete, plus a CHANGELOG entry — so per first-come I merged that one a few minutes ago. I don't want your work to go to waste, because frankly the *thinking* in this PR is a level above a plain delete and the project is better with it in the tree:

- Your reasoning for **not** wrapping `base64.b85encode` — that the alphabet here is deliberately in ASCII order so encoded integers sort lexicographically, and the stdlib alphabet isn't — is exactly right, and it's the kind of thing that gets "helpfully refactored" into a bug two years later by someone who doesn't know. That belongs in the module docstring, which you added.
- Your test suite is genuinely better than what landed: `test_alphabet_is_in_ascii_order`, `test_encoding_preserves_order`, and especially `test_sortable_int_to_text_preserves_order` (exercising the invariant *through* `whoosh2`'s own API) lock down the property the module actually exists for, not just a roundtrip.

So here's what I'd love, if you're up for it: a small follow-up PR on top of current `main` that (1) adds your explanatory module docstring and (2) replaces `tests/test_base85.py` with your richer property-based version. It'll merge cleanly and you'll have your own landed contribution — and I'll credit you in the CHANGELOG. Ping me here or on the new PR and I'll fast-track the review.

Closing this one only so the follow-up has a clean base — the credit is very much still yours. Thank you for taking the time to explain the *why*; that's exactly the instinct this project needs. 🙏

PR #143, opened 2026-08-25
Document why base85's alphabet is ordered, and test the invariant

chmm195:
Follow-up to #142 as you suggested, rebased on current `main` with #141 already in. Thanks for the generous writeup — @SAY-5 got there first and fair is fair.

Two pieces, exactly the two you asked for.

### 1. The docstring

It still said "generic base85 encoding and decoding functions", which stopped being true once the byte codec went, and gave no hint that the alphabet is deliberately ordered. The explanation now sits directly above the code someone would edit:

> Unlike the standard ascii85/base85 alphabets, the character set here is in ASCII order, so encoded strings sort in the same order as the integers they encode. `whoosh.codec.whoosh2`'s `sortable_int_to_text`/`sortable_long_to_text` rely on that property, so the alphabet must not be swapped for the one in the stdlib `base64` module […] substituting it would silently break sorted-order lookups on existing indexes.

I added the fixed-width half too, since it is the other reason the encoding sorts: without it a shorter string could compare less than a longer one just for being shorter.

### 2. The tests

Kept your `test_integer_roundtrip` (widened slightly) and **kept `test_byte_codec_removed` from #141** rather than replacing the file wholesale — you said "replaces", but a guard against the dead codec being reintroduced is worth more than the four lines it costs, and it is testing something none of my cases do. Say the word if you would rather it go.

Added on top:

| Test | Invariant |
| --- | --- |
| `test_alphabet_is_in_ascii_order` | ordered, 85 unique chars — the reason `base64.b85` cannot be substituted |
| `test_encoding_is_fixed_width` | 5 / 10 chars regardless of magnitude |
| `test_encoding_preserves_order` / `_long` | encoded order matches integer order |
| `test_adjacent_values_stay_adjacent_in_order` | 84/85/86 across the carry, which random sampling can miss |
| `test_sortable_int_to_text_preserves_order` / `_long` | the same property through `whoosh2`'s API, which is what actually depends on it |

The long-width order test uses `random.randrange` rather than `random.sample`: `range(85 ** 10)` overflows `ssize_t` when `sample` takes its `len()`.

### CHANGELOG

Added an `[Unreleased]` entry under Added/Changed in your existing style. I put my own attribution in it because you offered — please just edit or drop it if you would rather word that yourself.

### Verification

- `pytest tests/` — **871 passed**, 8 skipped (platform/optional-dep skips, unchanged from `main`)
- `pytest tests/test_base85.py` — 9 passed
- `ruff check` / `ruff format --check` — clean on both changed files

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,18 @@ All notable changes to this project are documented here. This project follows
 
 ## [Unreleased]
 
+### Added
+- Property-based tests for `whoosh.support.base85` covering the invariants the
+  module exists for: the alphabet being in ASCII order, fixed-width output, and
+  encoded values sorting in the same order as the integers they encode -- the
+  last asserted both directly and through `whoosh2`'s `sortable_int_to_text`
+  and `sortable_long_to_text` (#139). Thanks @chmm195.
+
+### Changed
+- `whoosh.support.base85`'s module docstring now explains why the alphabet is
+  in ASCII order and why it must not be replaced with `base64.b85encode`, whose
+  alphabet is unordered (#139). Thanks @chmm195.
+
 ## [3.47.0] - 2026-08-25
 
 ### Fixed
--- a/src/whoosh/support/base85.py
+++ b/src/whoosh/support/base85.py
@@ -1,7 +1,18 @@
 """
-This module contains generic base85 encoding and decoding functions. The
-whoosh.util.numeric module contains faster variants for encoding and
-decoding integers.
+This module contains base85 encoding and decoding functions for integers.
+The whoosh.util.numeric module contains faster variants.
+
+Unlike the standard ascii85/base85 alphabets, the character set here is in
+ASCII order, so encoded strings sort in the same order as the integers they
+encode. whoosh.codec.whoosh2's sortable_int_to_text/sortable_long_to_text
+rely on that property, so the alphabet must not be swapped for the one in
+the stdlib base64 module -- base64.b85encode's alphabet is not ordered, and
+substituting it would silently break sorted-order lookups on existing
+indexes.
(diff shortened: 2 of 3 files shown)


priya-sundaram-dev (approved):
This is exactly what I hoped the follow-up would look like — thank you for coming back for it, @chmm195.

A few things I want to call out specifically:

- **Keeping `test_byte_codec_removed`** from #141 rather than overwriting it was the right call. You read my "replaces" more carefully than I wrote it: the reintroduction guard and your invariant tests protect different things, and both are worth having. Please keep it.
- **The docstring placement** — putting the *why* directly above the code someone would edit, with the concrete failure mode ("silently break sorted-order lookups on existing indexes") — is the kind of comment that actually stops a future refactor. `base64.b85encode` is a genuinely tempting substitution and now there's a signpost.
- `test_adjacent_values_stay_adjacent_in_order` across the 84/85/86 carry, and asserting the property **through `sortable_int_to_text`/`sortable_long_to_text`** (the real dependents), are both testing something the roundtrip never could. And thank you for the `randrange` vs `sample` note on the `ssize_t` overflow — that would have bitten someone.

Attribution stays as you wrote it; you earned it. Verified locally: `test_base85.py` 9 passed, ruff + format clean. Approving now — will merge once the matrix goes green.

priya-sundaram-dev:
Merged — thanks again @chmm195. Two merged PRs now (#141 chain led here); really glad you came back for the follow-up. The base85 module is in genuinely better shape than it was under any prior maintainer: dead codec gone, the load-bearing sort invariant is now pinned by tests *and* documented at the point of temptation. That is a clean win. Hope to see you on the next one. 🙏

PR #145, opened 2026-08-25
Stricter tests after: Ruff rules E for pycodestyle errors

cclauss:
## Summary

* #138 was a bit too brisk and removed some strictness in tests where we want to assert that a value `is True` not just that it is truthy.

## Checklist

### DIFF
--- a/tests/test_codecs.py
+++ b/tests/test_codecs.py
@@ -747,7 +747,7 @@ def test_add_reader_to_index():
     writer.add_reader(reader)
 
     # Assert that the reader was added to the index
-    assert writer._added
+    assert writer._added is True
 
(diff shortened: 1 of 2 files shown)


priya-sundaram-dev (approved):
Thanks @cclauss — you're right that #138 was too brisk here. Asserting `is True`/`is False` on these boolean flags (`startexcl`, `endexcl`, `constantscore`, `_added`) is exactly the right level of strictness for these tests: it catches a regression where the value drifts to something truthy-but-not-`bool`. All 14 CI jobs green (3.10–3.15 + free-threaded, plus lint/mypy/pre-commit). Merging.

Issue #146, opened 2026-08-25
Free-Threading: Add `pytest-run-parallel` to some test runs in CI


cclauss:
### What happened?

Python Free-Threading Guide is https://py-free-threading.github.io
* https://py-free-threading.github.io/testing

Let's add [`pytest-run-parallel`](https://github.com/Quansight-Labs/pytest-run-parallel) to some of our pytest runs in CI because we have some tests that fail.

### Minimal reproduction

Caution: This runs for a long time even on my M1 MacBook Pro.
Using [`uv`](https://docs.astral.sh/uv) I ran:
```bash
uvx --with=pytest-run-parallel --with=. pytest --iterations=8 --parallel-threads=auto \
    --ignore=tests/test_automata.py \
    --ignore=tests/test_bump_version_script.py \
    --ignore=tests/test_cli.py \
    --ignore=tests/test_codecs.py \
    --ignore=tests/test_collector.py \
    --ignore=tests/test_writing.py
```
Without the `--ignore`s tests will fail or hang.  Indivvidual tests can be run like:
% `uvx --with=pytest-run-parallel --with=. pytest --iterations=8 --parallel-threads=auto tests/test_cli.py`

### Whoosh version

This repo's `main` branch

### Python version and OS

Python 3.14 on the most current macOS

priya-sundaram-dev:
Thanks for opening this, @cclauss — free-threading readiness is exactly the kind of modernization this library should take seriously. I dug into each failing group; here's the map so we can decide the CI shape together.

**Good news first:** the core is thread-safe. `test_queries`, `test_parsing`, `test_analysis`, and `test_automata` all pass clean at `--parallel-threads=auto --iterations`. No data races in query parsing, matching, analysis, or the FSA/FST code.

The failures fall into two buckets:

**1. A genuine bug (fixed in #147).** `TimeLimitCollector` armed `signal.SIGALRM` unconditionally, which raises `ValueError: signal only works in the main thread` off the main thread. That's a real-world bug for threaded web servers, not just a test artifact — so I fixed it: arm the alarm only on the main thread and fall back to the existing `threading.Timer` elsewhere (the limit is still enforced). This clears 2 of the 3 `test_collector.py` errors.

**2. Test-harness artifacts, not core races:**
[...]

PR #150, opened 2026-08-25
ci: install CI/docs/publish deps with uv (keep setup-python provisioning)

### LINKED ISSUE #149: Let's upgrade from Python pip to Astral uv
### What problem would this solve?

Modern Python projects are increasingly using [`uv`](https://docs.astral.sh/uv) for its speed, flexibility, and ease of use.

### Proposed solution

Use uv to install and test instead of pip.
Use https://github.com/astral-sh/setup-uv instead of https://github.com/actions/setup-python
Use https://github.com/astral-sh/uv-pre-commit to ensure our lockfile is up to date.
Modify our README to show users how `uv` can save them time and effort.

### Alternatives considered
I could have gone kitesurfing.

priya-sundaram-dev:
Implements the two pieces of #149 we agreed on, keeping the parts that fit a library.

## What changed
- **`uv` for installs in CI.** Swapped `python -m pip install …` → `uv pip install --system …` in the test, docs (`pages.yml`), and publish (`publish.yml`) workflows. `astral-sh/setup-uv` (pinned by SHA, already used by the lint/pre-commit jobs) is added to each job that installs.
- **`actions/setup-python` kept for provisioning.** It still installs the interpreters with `allow-prereleases: true`, so the 3.15 RC row and the free-threaded `3.14t`/`3.15t` rows — which have caught real threading races (#146, #148) — are preserved. Only the *install step* moved to `uv`.
- **README** gains a `uv pip install whoosh3` quickstart next to `pip`.

## What we deliberately left out
Per the discussion in #149, no committed `uv.lock` / `uv-pre-commit`. Whoosh is a library and should keep floating within its declared version ranges rather than pinning one resolution; a lockfile users never see would add churn without protecting them.

Thanks @cclauss for the push toward modern tooling. Closes #149.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -55,11 +55,11 @@ jobs:
         with:
           python-version: ${{ matrix.python-version }}
           allow-prereleases: true
+      - uses: astral-sh/setup-uv@20cfd1bf945f4377ade1205e4dbc17946fc9a30d # v10.0.1
       - name: Install
         run: |
-          python -m pip install --upgrade pip
-          python -m pip install -e "${{ matrix.extras }}"
-          python -m pip install pytest
+          uv pip install --system -e "${{ matrix.extras }}"
+          uv pip install --system pytest
       - name: Run tests
         run: python -m pytest tests/ -q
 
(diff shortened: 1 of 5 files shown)


cclauss (approved):
This is impressive!  I really like it, but...

We are using uv as a way to run pip while we could be adopting the uv workflow more completely, like:
In `pypproject.toml` adding `optional-dependencies.pytest_run_parallel` and using
uv build
uv check
uv format
uv publish

However, perhaps what I suggest is too big of a leap for the initial uv pull request.  I approve this pull request, and we can merge it as is to get the migration started, but in future PRs, let's adopt more of the uv workflow.

priya-sundaram-dev:
Thanks for the thoughtful review, @cclauss — and you're right that this PR only uses uv as a faster `pip`, leaving the fuller uv workflow (`uv build`, trusted-publishing via `uv publish`, a dev dependency-group for the test tooling) on the table. That was deliberate for a first migration, but it shouldn't get lost, so I've captured the follow-ups in #151 with a read on each and where the sharp edges are (the `uv publish` slice pairs nicely with the existing Trusted Publishing TODO). Grab any slice you like there and I'll assign it.

PR #155, opened 2026-08-26
ty: fix no-matching-overload and un-ignore in pyproject.toml

SantiagoDaleffe:
## Summary

This PR resolves the remaining `no-matching-overload` violation to help clear `pyproject.toml`.

Fixed one instance in `examples/rag_retriever.py` by replacing `key=scores.get` with `key=lambda k: scores[k]`. The type checker struggled with `dict.get` because its signature implies an `Optional` return type (which isn't strictly comparable for `sorted`), whereas direct indexing guarantees the correct type.

The rule has been successfully un-ignored in `pyproject.toml`.

## Related issues

References #121

## Checklist

### DIFF
--- a/examples/rag_retriever.py
+++ b/examples/rag_retriever.py
@@ -152,7 +152,7 @@ def reciprocal_rank_fusion(rankings, k=60):
     for ranking in rankings:
         for rank, doc_id in enumerate(ranking):
             scores[doc_id] = scores.get(doc_id, 0.0) + 1.0 / (k + rank + 1)
-    return sorted(scores, key=scores.get, reverse=True)
+    return sorted(scores, key=lambda k: scores[k], reverse=True)
 
 
 def hybrid_search(ix, query, k=10):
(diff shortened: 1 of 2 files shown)


priya-sundaram-dev (approved):
Thanks so much for this, @SantiagoDaleffe — and welcome! 🎉

This is a clean, well-reasoned fix. Your diagnosis is exactly right: `dict.get`'s overloads make the return `Optional` in the checker's eyes (since it can't prove a default is always supplied), so it can't be used directly as a `sorted` key without tripping `no-matching-overload`. Switching to `key=lambda k: scores[k]` guarantees the value type, and since every key in `scores` is present it's functionally identical.

I ran `ty check` locally with the rule un-ignored and it comes back **All checks passed!** — so the stale `# 2 instances` comment was over-counting (the other one had already been cleared elsewhere). All 16 CI jobs are green too. Merging now. Really appreciate you picking up an item from the #121 tracking issue — this is precisely the kind of incremental typing cleanup it was meant to invite.

PR #156, opened 2026-08-27
Ty: un-ignore invalid-return-type rule

SantiagoDaleffe:
## Summary

This PR addresses the 28 violations of `invalid-return-type`, allowing us to un-ignore the rule in `pyproject.toml`. 

The fixes fall into a few architectural categories:
1. **AlwaysTruthy in `__eq__` methods**: Fixed multiple `__eq__` implementations in `analyzers.py` and `filters.py` that were returning `other and ...` (which evaluates to `None` if `other` is `None`, violating the `bool` return type). Replaced with `type(self) is type(other)` to guarantee a strict boolean return.
2. **Tokenizer vs Analyzer Inheritance**: Updated the return types of factory functions like `IDAnalyzer` and `KeywordAnalyzer`. They were annotated to return `Analyzer` but conditionally returned `Tokenizer`s. Changed the return type to their common base class, `Composable`.
3. **Union & Optional strictness**: 
   - Added `| None` to `fields.py` where a function explicitly returned `None`.
   - Updated `FacetMap.map` in `sorting.py` to correctly reflect that it can return the uninstantiated class `type[FacetMap]`.
4. **Casting**: Added `typing.cast("qcore.Query", ...)` in `RangeMixin.merge` since the linter couldn't infer the final class type from `self.__class__`.

The rule is now successfully un-ignored and fully enforced.

## Related issues

References #121

## Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -192,7 +192,6 @@ rules.invalid-argument-type = "ignore"  # 30 instances
 rules.invalid-assignment = "ignore"  # 21 instances
 rules.invalid-ignore-comment = "ignore"  # 1 instance
 rules.invalid-method-override = "ignore"  # 131 instances
-rules.invalid-return-type = "ignore"  # 28 instances
 rules.missing-argument = "ignore"  # 12 instances
 rules.not-subscriptable = "ignore"  # 15 instances
 rules.too-many-positional-arguments = "ignore"  # 27 instances
--- a/src/whoosh/analysis/analyzers.py
+++ b/src/whoosh/analysis/analyzers.py
@@ -64,11 +64,7 @@ class Analyzer(Composable):
         return f"{self.__class__.__name__}()"
 
     def __eq__(self, other: object) -> bool:
-        return (
-            other
-            and self.__class__ is other.__class__
-            and self.__dict__ == other.__dict__
-        )
+        return type(self) is type(other) and self.__dict__ == other.__dict__
 
     def __call__(self, value: str, **kwargs: Any) -> Iterator[Token]:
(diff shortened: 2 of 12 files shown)


priya-sundaram-dev:
Merged — thank you, @SantiagoDaleffe! 🎉 This is a meaty one: 28 `invalid-return-type` violations cleared and the rule is now fully enforced.

A few things I especially appreciated on review:

- **The `__eq__` refactor to `type(self) is type(other)`** doesn't just satisfy the return-type rule — it fixes a latent correctness bug. The old `other and ...` form short-circuits on *any* falsy `other`, and several of our objects are falsy when empty (`CompositeAnalyzer` / `CompoundQuery` define `__len__`). So two equal empty composites — e.g. `And([]) == And([])` — used to come back non-`True`. Verified locally that it's now correct: `And([]) == And([])` → `True`, `Term(...) == None` → `False`, cross-type stays `False`.
- **Widening the `IDAnalyzer`/`KeywordAnalyzer`/`RegexAnalyzer` return types to `Composable`** is the honest annotation — those factories return `Tokenizer`s at runtime, so `Analyzer` was never accurate.
- Clean use of `cast(...)` in `RangeMixin.merge` and `Query | None` in `FieldType.parse_range` to match the documented contracts.

Verified before merge: full CI green (tests 3.10→3.15 incl. free-threaded, ruff, pre-commit, mypy smoke), `ty check` → *All checks passed!* with the rule un-ignored, and the analysis/query/parsing/sorting/fields suites (220 tests) pass locally.

That's another rule off the tracking list. I'll update #121.

PR #159, opened 2026-08-29
typing: concrete parameter types where determinable (revised per review)

priya-sundaram-dev:
Revised after @cclauss's review. The original version enabled `ANN401` and swapped `Any` → `object` wholesale; as you noted, that helps neither readers nor type checkers, since everything is an `object`.

Reworked to do the analysis instead — annotate the parameters whose type the code and docstrings already pin down, and keep `Any` where the value is genuinely polymorphic:

**Now concretely typed:**
- `Query.matcher(context=...)` → `SearchContext | None`
- `FieldType.sortable_terms(ixreader=...)` and `Expander(ixreader=...)` → `IndexReader`
- `BitSet` / `OnDiskBitSet` on-disk methods (`dbfile=...`) → `StructFile` (already documented as such in the docstrings)

The concrete `ixreader` type also surfaced a latent inaccuracy: the base `FieldType.sortable_terms` returns `reader.lexicon(...)`, which is `Iterable[bytes]`, not the annotated `Iterator[bytes]` — corrected.

**Deliberately kept `Any`:** term text and stored field values. `Term('price', 50)` and stored `NUMERIC` values are legitimately non-`str` (the test suite constructs `Term`s with ints), so `str` would be too narrow and `Any` is the honest type — exactly the "might be impossible to determine" case you described. For the same reason I dropped the blanket ANN401 gate; enforcing it would just push those honest `Any`s back to `object` or a wall of `# noqa`.

Green locally: `ruff check`, `ruff format --check`, `ty check`, `mypy` smoke, `pytest` (883 passed).

### DIFF
--- a/src/whoosh/classify.py
+++ b/src/whoosh/classify.py
@@ -39,6 +39,8 @@ from typing import TYPE_CHECKING, Any
 if TYPE_CHECKING:
     from collections.abc import Iterable
 
+    from whoosh.reading import IndexReader
+
 
 # Expansion models
 
@@ -117,7 +119,7 @@ class Expander:
 
     def __init__(
         self,
-        ixreader: Any,
+        ixreader: IndexReader,
         fieldname: str,
         model: type[ExpansionModel] | ExpansionModel = Bo1Model,
     ) -> None:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,18 @@ All notable changes to this project are documented here. This project follows
   targeted `# ty: ignore[unknown-argument]` on mixin `self.__class__(...)`
   re-instantiations where the concrete constructor can't be known statically.
   Thanks to @SantiagoDaleffe.
+- Replaced a few opaque `Any` parameter annotations with the concrete types the
+  code and docstrings already document, so editors and type checkers can help
+  callers: `Query.matcher(context=...)` is now `SearchContext | None`,
+  `FieldType.sortable_terms(ixreader=...)` and `Expander(ixreader=...)` take an
+  `IndexReader`, and `BitSet`/`OnDiskBitSet`'s on-disk methods take a
+  `StructFile`. The concrete `ixreader` type also surfaced that the base
+  `FieldType.sortable_terms` returns an `Iterable[bytes]` (from
+  `reader.lexicon()`), not an `Iterator[bytes]`; the annotation is corrected.
+  Genuinely polymorphic boundaries — term text and stored field values, which
+  legitimately accept `str`, numbers, `bytes`, … — keep `Any`, since a narrower
+  type there would be wrong. Thanks to @cclauss for the review that prompted
+  this analysis.
 
 ## [3.49.0] - 2026-08-28
 
--- a/src/whoosh/fields.py
+++ b/src/whoosh/fields.py
@@ -50,6 +50,7 @@ if TYPE_CHECKING:
     from whoosh.analysis import Analyzer, Token
     from whoosh.formats import Format
     from whoosh.query import Query
+    from whoosh.reading import IndexReader
 
 from whoosh import analysis, columns, formats
 from whoosh.system import emptybytes, pack_byte
@@ -290,7 +291,7 @@ class FieldType:
         else:
             self.column_type = None
 
-    def sortable_terms(self, ixreader: Any, fieldname: str) -> Iterator[bytes]:
+    def sortable_terms(self, ixreader: IndexReader, fieldname: str) -> Iterable[bytes]:
         """
         Returns an iterator of the "sortable" tokens in the given reader and
         field. These values can be used for sorting. The default implementation
@@ -900,7 +901,7 @@ class NUMERIC(FieldType):
             fieldname, start, end, startexcl, endexcl, boost=boost
         )
 
-    def sortable_terms(self, ixreader: Any, fieldname: str) -> Iterator[bytes]:
+    def sortable_terms(self, ixreader: IndexReader, fieldname: str) -> Iterator[bytes]:
         zero = b"\x00"
         for token in ixreader.lexicon(fieldname):
             if token[0:1] != zero:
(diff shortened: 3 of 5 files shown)


src/whoosh/query/terms.py
```diff
@@ -278,7 +278,7 @@ class PatternQuery(MultiTerm):
     def __init__(
         self,
         fieldname: str,
-        text: Any,
+        text: object,
```
cclauss:
```suggestion
        text: str,
```

src/whoosh/query/terms.py
```diff
@@ -425,7 +425,7 @@ def __str__(self) -> str:
         return f'{self.fieldname}:r"{self.text}"'
 
     def _get_pattern(self) -> str:
-        return self.text
+        return cast("str", self.text)
```
cclauss:
```suggestion
        return self.text
```

src/whoosh/query/terms.py
```diff
@@ -487,7 +487,7 @@ class FuzzyTerm(ExpandingTerm):
     def __init__(
         self,
         fieldname: str,
-        text: Any,
+        text: object,
```
cclauss:
```suggestion
        text: str,
```

src/whoosh/query/terms.py
```diff
@@ -557,7 +557,7 @@ def _btexts(self, ixreader: IndexReader) -> Iterable[Any]:
             self.fieldname, self.text, self.maxdist, prefix=self.prefixlength
         )
 
-    def replace(self, fieldname: str, oldtext: Any, newtext: Any) -> qcore.Query:
+    def replace(self, fieldname: str, oldtext: object, newtext: object) -> qcore.Query:
```
cclauss:
```suggestion
    def replace(self, fieldname: str, oldtext: str, newtext: str) -> qcore.Query:
```

src/whoosh/query/terms.py
```diff
@@ -569,7 +569,7 @@ class Variations(ExpandingTerm):
     given word in the same field.
     """
 
-    def __init__(self, fieldname: str, text: Any, boost: float = 1.0):
+    def __init__(self, fieldname: str, text: object, boost: float = 1.0):
```
cclauss:
```suggestion
    def __init__(self, fieldname: str, text: str, boost: float = 1.0):
```

src/whoosh/query/terms.py
```diff
@@ -607,7 +607,7 @@ def _btexts(self, ixreader: IndexReader) -> Iterator[bytes]:
     def __str__(self) -> str:
         return f"{self.fieldname}:<{self.text}>"
 
-    def replace(self, fieldname: str, oldtext: Any, newtext: Any) -> qcore.Query:
+    def replace(self, fieldname: str, oldtext: object, newtext: object) -> qcore.Query:
```
cclauss:
```suggestion
    def replace(self, fieldname: str, oldtext: str, newtext: str) -> qcore.Query:
```

src/whoosh/classify.py
```diff
@@ -173,7 +173,7 @@ def add_document(self, docnum: int) -> None:
                 % (self.fieldname, docnum)
             )
 
-    def add_text(self, string: Any) -> None:
+    def add_text(self, string: object) -> None:
```
cclauss:
```suggestion
    def add_text(self, string: str) -> None:
```

cclauss (changes requested):
As it currently stands, this pull request can be closed without merging any changes.

We add type annotations to Python code so that contributors and type checkers can know what data types are expected.

There are naming conventions that let us assume that variable names like `string`, `text`, `oldtext`, and `newtext' should have type `str`.  Similarly, variable names like `count`, `number`, or `num` should have type `int`.  We annotate them like that, and if `ty` or `mypy` tells us we are wrong, we drop back to `Any`.

Given that everything in Python is an `object`, the current changes in this PR do not help contributors or type checkers to know what data types are expected.  When trying to fix `ruff rule ANN401`, we should determine which data type(s) make sense given what the code is doing.  It requires analysis, and it might be impossible to determine.  If we are not improving the state of the code for the reader or type checkers, then we should just stick with `Any`.

priya-sundaram-dev:
Thanks @cclauss — you're right, `object` was no improvement over `Any`. I've reworked this to do the analysis you described rather than a mechanical substitution.

I dug into the specific spots you flagged. Interesting result: `Term`'s `text` (and the `stored field value` in `Expander.add_text`) can't safely be `str` — the test suite constructs `Term('price', 50)` / `Term('seats', 260)` against NUMERIC fields, and `field.to_bytes()` handles the conversion, so those are legitimately polymorphic. That's the "impossible to determine" case, so I left them `Any`.

Where the type *is* determinable I annotated it: `context` → `SearchContext | None`, `ixreader` → `IndexReader`, `dbfile` → `StructFile` (the docstrings already said as much). The concrete `ixreader` type even caught that `sortable_terms` returns `Iterable[bytes]`, not `Iterator[bytes]`.

I also dropped the blanket ANN401 enable — the remaining violations are the genuinely-dynamic value boundaries where you advised sticking with `Any`. PR description updated; all checks green. Happy to keep going module-by-module if you'd like.

cclauss (approved):
Nice work!  Thanks for doing this.  They call it gradual typing because it takes a lot of detective work and trial and error to discover the right types in a preexisting codebase.

We should always encourage contributors to add types as they write new code.

priya-sundaram-dev:
Thanks for the thorough review, @cclauss — the `Term('price', 50)` case was a great reminder that the honest answer sometimes really is `Any`. Merged. Appreciate you pushing for the concrete-type detective work rather than a mechanical substitution.

cclauss:
I would have assumed that a `price` or `cost` was a currency (float) to hold dollars and cents.

priya-sundaram-dev:
Ha, exactly — a real `price` field would almost certainly be a `NUMERIC(float)` for dollars and cents. That's really the whole point: the term *text* passed to `Term(fieldname, value)` takes whatever Python type matches the field (int, float, `datetime`, bytes, str), and `field.to_bytes()` does the conversion. So there's no single concrete annotation that's honest across text/numeric/datetime fields — `Any` is the truthful one there, even though it's tempting to tighten it. Thanks again for pushing on this; the codebase is better typed for it.

PR #179, opened 2026-09-04
Add type hints for whoosh.util.varints

asarakhatun17-lgtm:
## Summary

Added type hints to src/whoosh/util/varints.py and updated CHANGELOG.
## Checklist

- [ ✔️] Tests pass locally (`pytest tests/`)
- [ ✔️] Added/updated tests for the change where it makes sense
- [ ✔️] Updated docs / CHANGELOG if user-facing
- [ ✔️] I have read the [CONTRIBUTING](https://github.com/priya-sundaram-dev/whoosh/blob/main/CONTRIBUTING.md) guide

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,8 +8,6 @@ All notable changes to this project are documented here. This project follows
 
 ### Changed
 
-- Add type hints for whoosh.util.varints
-
 - Typing: added parameter and return annotations to all public functions in
   `whoosh.util.text` (`byte`, `first_diff`, `prefix_encode`,
   `prefix_encode_all`, `prefix_decode_all`, `natural_key`, `rcompile`)
@@ -27,7 +25,7 @@ All notable changes to this project are documented here. This project follows
   bound resolves straight to a concrete instant — for example
   `added:[noon TO now]` or `added:[noon TO -1 week]` — no longer crash the
   parser with `AttributeError: 'datetime.datetime' object has no attribute
-'ceil'`. The range-disambiguation step compared `start.floor().time()` with
+  'ceil'`. The range-disambiguation step compared `start.floor().time()` with
   `end.ceil().time()` by calling the `adatetime` methods directly, but a bound
   that resolves to a plain `datetime` (like `now` or a relative offset) has no
   such methods. Both sides now go through the module-level `floor()`/`ceil()`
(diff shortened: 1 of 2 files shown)


priya-sundaram-dev:
Thank you for the PR, @asarakhatun17-lgtm — and welcome! 🎉 The annotations on `varints.py` itself are spot on: `varint`/`_varint` → `bytes`, the integer params, and `read_varint(readfn: Callable[[int], bytes]) -> int` are all exactly right, and pulling `Callable` from `collections.abc` is the modern choice. I approved the CI run so the checks can execute.

Two small things and this is good to merge:

1. **CHANGELOG.md picked up a lot of unrelated reformatting.** It looks like an editor with a Markdown auto-formatter ran over the whole file — the diff touches ~40 lines it shouldn't (blank lines inserted after `### Fixed`/`### Added` headings, `*before*` → `_before_` emphasis swaps, and a few continuation lines lost their leading indentation, e.g. the `'ceil'.`, `(4 given)`, and `"42")` lines now start at column 0, which changes how they render). Could you revert everything in `CHANGELOG.md` except your one new line? The cleanest way:

   ```bash
   git checkout main -- CHANGELOG.md
   ```
   then re-add just your entry under `## [Unreleased]` → `### Changed`:
   ```
   - Add type hints for whoosh.util.varints
   ```
   Keeping the diff to only the lines you meant to change makes review trivial and avoids churn for the next person reading the history.

2. **Tiny stray space** in the first signature: `def _varint(i: int ) -> bytes:` → `def _varint(i: int) -> bytes:`. (The `ruff format` CI step will flag this, so fixing it now keeps the checks green.)

One optional FYI, not a blocker: `varint_to_int` has no callers left in the source tree and its body does `ord(vi[0])`, which actually can't accept `bytes` at runtime (in Py3, `vi[0]` is already an `int`). Your `vi: bytes` annotation documents the *intended* type fine — leave it as is; I just wanted you to know it's latent dead code, not something you need to fix here.

Nice, focused first contribution. Once the CHANGELOG is reverted to just your line and the stray space is gone, I'll merge. 🙌

priya-sundaram-dev:
Nicely done, @asarakhatun17-lgtm — thank you for reverting the CHANGELOG churn, that made this much easier to review. The annotations on `varints.py` are all correct. ✅

I've pushed one tiny tidy-up commit on top of yours (I hope you don't mind — it saves you a round-trip):

- Dropped a stray space in `def _varint(i: int )` → `(i: int)` so `ruff format`/pre-commit is happy.
- Moved your CHANGELOG line up under the `## [Unreleased]` → `### Changed` heading (it had landed at the very bottom of the file, below the "Older history…" note), and expanded it slightly to list the functions, matching the style of the neighbouring `util.text` entry.

The credit for the code is all yours. Re-running CI now — once it's green I'll squash-merge. Thanks for the clean contribution, and I hope you'll pick up another! 🎉

PR #188, opened 2026-09-11
ci: add PyPy 3.11 to the test matrix

deronnax:
## Summary
* Add a `pypy-3.11` row to the CI test matrix (base install only, like the free-threaded builds).
* Update the README Maintenance section to reflect that PyPy is now CI-tested.
* Fix `DFA.minimize()` to pick equivalence-class representatives deterministically (`min(part)` instead of `set.pop()`), which was failing on PyPy.

The README already claims PyPy support, but CI only exercised CPython. Optional integration extras (`langchain`, `llamaindex`, `mcp`) are skipped on PyPy because they don't reliably ship PyPy wheels — the pure-Python core is what matters here.

## Test plan
* [x] CI passes on fork PR (including `Test (Python pypy-3.11 / ubuntu-latest)` — 906 passed, 4 skipped).
* [ ] CI passes on upstream PR.


priya-sundaram-dev:
Thanks for this, @deronnax — this is a great contribution. Three well-scoped changes and the reasoning is spot on.

The `min(part)` fix is my favourite part: `set.pop()` on a set of integer states is order-nondeterministic (hash-seed / interpreter dependent), so `minimize()` could pick a different representative per run and per implementation. Picking the minimum makes the output stable, which is exactly what let the DFA-equality assertions pass reliably on PyPy. I reviewed it locally and confirmed `is initial` identity still holds — `min()` returns the actual element object from the set, so nothing changes for CPython, and `tests/test_automata.py` + `tests/test_spelling.py` stay green here.

The README already promised PyPy support, so wiring a `pypy-3.11` base-install row into the matrix (and skipping the optional integration extras that don't ship PyPy wheels) is the right call — it makes the promise real instead of aspirational.

I've approved the workflow run so CI exercises the new PyPy row on this PR. Assuming it comes back green like it did on your fork (906 passed / 4 skipped), I'll merge. Really appreciate you digging into the FSA determinism rather than just papering over it. 🙏

priya-sundaram-dev:
Merged — CI came back fully green including the new `pypy-3.11` row (906 passed / 4 skipped, matching your fork). Thanks again @deronnax; the FSA determinism fix and PyPy coverage are both landing in the next release, and you'll be credited in the changelog. 🎉

Issue #193, opened 2026-09-12
Type hints for whoosh.filedb.structfile (good first issue)


priya-sundaram-dev:
### Summary

Add type hints to `src/whoosh/filedb/structfile.py`. `StructFile` is a thin wrapper around a binary file object that adds typed `read_*`/`write_*` helpers (ints, floats, varints, strings, arrays, pickles). The signatures are almost self-documenting — a great first contribution, same shape as the recently-merged `util.varints` (#179), `util.text` (#175), `util.numeric` (#172) and `util.versions` (#186) typing PRs.

### Why

We're gradually typing the codebase so `ty` can catch real bugs (tracker: #121). `structfile.py` is foundational I/O but its methods have very clear input/output types, so the annotations are unambiguous and easy to review.

### What to do

Annotate the methods on `StructFile` (and the tiny helpers). The `write_*` methods return `None`; the `read_*` methods return the value they read. Suggested signatures for the main groups:

```python
from __future__ import annotations
from array import array
from typing import Any, BinaryIO, Callable

class StructFile:
    def __init__(self, fileobj: BinaryIO, name: str | None = None,
                 onclose: Callable[[StructFile], None] | None = None) -> None: ...

    # fixed-width numbers
    def write_int(self, n: int) -> None: ...
    def read_int(self) -> int: ...
    def write_uint(self, n: int) -> None: ...
    def read_uint(self) -> int: ...
    def write_ushort(self, n: int) -> None: ...
    def read_ushort(self) -> int: ...
    def write_long(self, n: int) -> None: ...
    def read_long(self) -> int: ...
    def write_float(self, n: float) -> None: ...
    def read_float(self) -> float: ...
    def write_byte(self, n: int) -> None: ...
    def read_byte(self) -> int: ...

    # varints
    def write_varint(self, i: int) -> None: ...
    def read_varint(self) -> int: ...

    # length-prefixed bytestrings (Whoosh "strings" are bytes on disk)
    def write_string(self, s: bytes) -> None: ...
    def read_string(self) -> bytes: ...

    # arrays and pickles
    def write_array(self, arry: array) -> None: ...
    def read_array(self, typecode: str, length: int) -> array: ...
    def write_pickle(self, obj: Any, protocol: int = -1) -> None: ...
    def read_pickle(self) -> Any: ...
```

Notes:
[...]

Issue #186, opened 2026-09-05
Type hints for whoosh.util.versions (good first issue)


priya-sundaram-dev:
### Summary

Add type hints to `src/whoosh/util/versions.py`. It's a small, self-contained module (version-number parsing and comparison), so it's a great first contribution — same shape as the recently-merged `util.varints` (#179), `util.text` (#175), and `util.numeric` (#172) typing PRs.

### Why

We're gradually typing the codebase so `ty` can catch real bugs (tracker: #121). `versions.py` has no external dependencies beyond `whoosh.util.text.rcompile`, so the annotations are unambiguous and easy to review.

### What to do

Annotate the methods on `BaseVersion` and `SimpleVersion`. Suggested signatures:

```python
from __future__ import annotations
from typing import ClassVar

class BaseVersion:
    @classmethod
    def parse(cls, text: str) -> "BaseVersion": ...
    def __repr__(self) -> str: ...
    def tuple(self) -> tuple: ...
    def __eq__(self, other: object) -> bool: ...
    def __lt__(self, other: object) -> bool: ...
    # ... __gt__/__ge__/__le__/__ne__ same as __eq__/__lt__

class SimpleVersion(BaseVersion):
    def __init__(self, major: int = 1, minor: int = 0, release: int = 0,
                 ex: str = "z", exnum: int = 0) -> None: ...
    def to_int(self) -> int: ...
    @classmethod
    def from_int(cls, n: int) -> "SimpleVersion": ...
```

Notes:
- The six rich-comparison dunders (`__eq__`, `__lt__`, `__gt__`, `__ge__`, `__le__`, `__ne__`) all take `other: object` and return `bool` — matches the `object.__eq__` signature so `ty` stays happy.
- `parse`/`from_int` are `@classmethod`s that build and return an instance; a forward-reference string return type is fine.
- Class-level attributes like `_ex_bits`/`_bits_ex`/`_parts` can optionally get `ClassVar[...]` annotations, but the priority is the method signatures.
- No behavior changes — annotations only.

### How to verify

```bash
uv run pytest tests/            # existing tests still pass
uv run ty check src/whoosh/util/versions.py   # no new errors
```

Then add a line under `[Unreleased]` in `CHANGELOG.md`. Comment here if you'd like it assigned — happy to help if anything's unclear.

`good first issue` · `typing`

11solarflash11:
Stop

priya-sundaram-dev:
[...]

PR #175, opened 2026-09-03
feat(typing): add type annotations to whoosh.util.text

webdevsamran:
Fixes #173 (part of #121)

Adds parameter and return type annotations to all public functions in \src/whoosh/util/text.py\:
- \yte(num: int) -> bytes\
- \irst_diff(a: Sequence[Any], b: Sequence[Any]) -> int\
- \prefix_encode(a: str | bytes, b: str | bytes) -> bytes\
- \prefix_encode_all(ls: Iterable[str]) -> Iterator[bytes]\
- \prefix_decode_all(ls: Iterable[bytes]) -> Iterator[str]\
- \
atural_key(s: str) -> tuple[int | str, ...]\
- \
compile(pattern: str | re.Pattern[Any], flags: int = 0, verbose: bool = False) -> re.Pattern[Any]\
- Internal \_nkconv(i: str) -> int | str\

Also includes \rom __future__ import annotations\, defensive coercions in prefix encoding helpers, and records the change under \## [Unreleased]\ in \CHANGELOG.md\.

All local checks pass:
- \	y check src/whoosh/util/text.py\
- \
uff check src/whoosh/util/text.py\
- \
uff format --check src/whoosh/util/text.py\
- \pytest tests/test_sorting.py -q\ (37/37 passed)

PR #172, opened 2026-09-03
feat(typing): add type annotations to whoosh.util.numeric

webdevsamran:
﻿Fixes #171

Adds parameter and return type annotations to all public functions in `src/whoosh/util/numeric.py`: 
- `bits_required(maxnum: int) -> int`
- `typecode_required(maxnum: int) -> str`
- `max_value(bitcount: int) -> int`
- `bytes_for_bits(bitcount: int) -> int`
- `to_sortable(numtype: type[int | float], intsize: int, signed: bool, x: float) -> int`
- `from_sortable(numtype: type[int | float], intsize: int, signed: bool, x: int) -> int | float`
- `float_to_sortable_long(x: float, signed: bool) -> int`
- `sortable_long_to_float(x: int, signed: bool) -> float`
- `float_to_byte(value: float, mantissabits: int = 5, zeroexp: int = 2) -> bytes`
- `byte_to_float(b: bytes | int, mantissabits: int = 5, zeroexp: int = 2) -> float`
- `length_to_byte(length: int | None) -> int`

Also records the update under `## [Unreleased]` in `CHANGELOG.md`. No runtime behavior is changed.

Verified with `ruff check` and `ruff format` (all checks pass).

PR #194, opened 2026-09-13
feat(typing): add type annotations to whoosh.filedb.structfile (#193)

webdevsamran:
## Summary

Adds type annotations to ``whoosh.filedb.structfile`` as requested in #193. Same annotations-only pattern as the recently merged ``util.numeric`` (#172), ``util.text`` (#175), ``util.varints`` (#179), and ``util.versions`` (#186) PRs.

## What changed

**``src/whoosh/filedb/structfile.py``** � annotations only, zero behaviour changes:

- ``StructFile.__init__``: ``fileobj: BinaryIO``, ``name: str | None``, ``onclose: Callable[[StructFile], None] | None``
- All ``write_*`` methods: parameter types + ``-> None``
- All ``read_*`` methods: ``-> int``, ``-> float``, ``-> bytes``, or ``-> array`` as appropriate
- ``get_string2``/``get_string4``: ``-> tuple[bytes, int]``
- Pass-through wrappers (``read``, ``write``, ``seek``, ``tell``, ``truncate``, ``readline``): ``*args: Any, **kwargs: Any -> Any`` per issue spec
- ``BufferFile.__init__``: ``buf: bytes | bytearray | memoryview``
- ``BufferFile`` and ``ChecksumFile`` methods fully annotated
- ``Callable`` moved to ``collections.abc`` (ruff UP035)
- ``__exit__`` uses ``type[BaseException] | None``, ``BaseException | None``, ``types.TracebackType | None`` (ruff PYI036)

**``CHANGELOG.md``**: entry added under ``[Unreleased] ? Typing``.

## Verification

```
ruff check src/whoosh/filedb/structfile.py     # All checks passed
ruff format --check src/whoosh/filedb/structfile.py  # 1 file already formatted
python -m pytest tests/test_misc.py tests/test_reading.py -q  # 23 passed
```

## AI disclosure

Drafted with AI assistance (Gemini / Antigravity). All type signatures were derived directly from the method implementations and the suggested signatures in the issue body � no logic was altered.


priya-sundaram-dev (approved):
Thanks @webdevsamran — this is your third typing PR and it shows: clean, mechanical, behaviour-preserving annotations exactly matching the pattern from #172/#175/#179/#186. I re-ran everything locally against the checked-out branch:

- `ruff check` + `ruff format --check` → clean
- `python -m mypy` (the public-API smoke fixture) → `Success: no issues found`
- `pytest tests/test_misc.py tests/test_reading.py tests/test_writing.py` → 50 passed

A few things I especially liked: `from __future__ import annotations` so the `StructFile`/`BufferFile` forward refs resolve without quotes, `Callable` pulled from `collections.abc` (UP035), and the PYI036-correct `__exit__` triple (`type[BaseException] | None`, `BaseException | None`, `types.TracebackType | None`). The `*args: Any, **kwargs: Any -> Any` pass-throughs are the right call for the thin wrappers.

Approving. **One tiny nit** (not blocking): line 255 in `write_byte`'s docstring picked up a stray trailing backslash:

```python
        ``file.write(chr(n))``.\
```

Since it's an unescaped line-continuation in a plain docstring it silently merges the closing-quote line, so the rendered `__doc__` ends with `.` + 8 spaces instead of a newline — a small behaviour change that slips past the "annotations-only" intent. Easiest fix is to just drop the `\` so it reads ``` ``file.write(chr(n))``. ``` again.

Happy to squash that out myself when I merge so you don't have to round-trip — but if you'd rather push the one-char fix to keep your commit clean, go for it and I'll merge right after. Either way, thank you for another solid contribution to the typing effort (#121).

webdevsamran:
Good catch on the docstring line-continuation! Dropped the stray backslash in c53f10df so the commit history stays clean. Thanks for the quick review and approval!

priya-sundaram-dev:
Nice — history's clean now. One last thing before I merge: CI's `ty` hook (in Pre-commit) is failing, and it's actually a *good* catch that your annotations surfaced:

```
error[invalid-return-type]: Return type does not match returned value
   --> src/whoosh/filedb/structfile.py:102:16
101 |     def __str__(self) -> str:
102 |         return self._name
    |                ^^^^^^^^^^ expected `str`, found `str | None`
```

`_name` is `str | None`, so `__str__` can return `None`. That's a **pre-existing latent bug** your typing work uncovered — verified it locally:

```python
>>> str(StructFile(io.BytesIO(), name=None))
TypeError: __str__ returned non-string (type NoneType)
```

The minimal fix that keeps behavior for named files and satisfies `ty`:

```python
def __str__(self) -> str:
    return self._name if self._name is not None else ""
```

That's a one-liner and fully in the spirit of this PR (making the types honest). Want to push it, or would you prefer I squash it in on merge with credit to you? Either's fine by me — just let me know. Thanks again for the careful work here. 🙏

webdevsamran:
Great catch by `ty`! Pushed commit b809c53d with the fix (`return self._name if self._name is not None else ""`) along with a regression unit test in `tests/test_misc.py` verifying both `name=None` and `name="test"`. `ruff check`, `ruff format`, and all 51 tests pass cleanly.

priya-sundaram-dev:
Merged — thank you, @webdevsamran! 🎉 This is exactly the kind of contribution I love to see: the annotations are clean, and your typing work surfaced (and fixed) a real latent bug in `__str__` for unnamed `StructFile`s, complete with a regression test. Verified locally before merge: `ruff check`, `ruff format --check`, `ty check`, and the full test suite (908 passed) are all green.

If you enjoyed this, `whoosh.filedb` has a few more sibling modules that would benefit from the same treatment (`filestore.py`, `filetables.py`) — no pressure at all, but you would be very welcome. Thanks again for the careful, responsive work.

